PR #4, opened 2015-02-20
xml_root is not defined

renkun-ken:
`xml_root` is not defined. Use other examples to demonstrate the ability of parsing incomplete html document.


hadley:
Thanks!

PR #18, opened 2015-03-11
New xml_save method for serializing xml DOM to char vector.

sjp:
Hi,

I was playing around with the `xml2` package and figured it would be good if we could serialise the DOM back to a character vector.

What I've implemented is nothing major, just defaults to what already exist for serialisation via the `print` methods.

It's missing a few features found in the `XML` package with its `saveXML` method but it should suffice for now.


hadley:
I wonder if this would be better as a method for `as.character()` or `format()`?

sjp:
Good idea.

I think I like the idea of an `as.character()` method better than `format()`, mostly because we're not ever going to be doing much formatting (besides indentation).

However, I guess an argument for `xml_save()` is that we do have the flexibility of adding arguments to the generic method.

What is your preference?

hadley:
`as.character()` has `...` so you can add arbitrary arguments. I think that suggests `as.character()` would be best

sjp:
Ok. I'll close this one off later and create a new PR for `as.character()`
methods.

PR #19, opened 2015-03-12
as.character methods for xml_document and xml_nodeset.

sjp:
As above.

Like the previous PR we don't have anything interesting going on in comparison to `XML::saveXML()`.

I left out documentation for these methods considering they're straightforward. If we decide to add a parameter later (e.g. indentation string) we can document it then.

Thanks.


hadley:
Thanks!

PR #21, opened 2015-04-10
Respect the indent parameter when printing xml_structure.

sjp:
As above. The indent parameter is now actually used when printing indentation for `xml_structure()` when previously it was always using 2 chars.

### DIFF
--- a/R/xml_structure.R
+++ b/R/xml_structure.R
@@ -13,15 +13,15 @@
 #' rproj <- read_html(system.file("extdata","r-project.html", package = "xml2"))
 #' xml_structure(rproj)
 #' xml_structure(xml_find_all(rproj, ".//p"))
-xml_structure <- function(x, indent = 0) {
+xml_structure <- function(x, indent = 2) {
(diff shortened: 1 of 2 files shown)


hadley:
Would you mind commenting on the `one_child` related change too?

sjp:
Sure.

When I was testing this out on some XML documents of my own, it often printed very wide, particularly when there are many attributes on an element with a single child.

Also a bit of an aesthetic choice too. When a single child was being printed alongside its parent it didn't make it clear to me that it was a child, rather than another parent. The intent was clearer (to me) when there was no differentiation between single and multiple children.

hadley:
Ok, makes sense - thanks! (I'm also working on something that will hopefully eventually work itself into an interactive tree style explorer in RStudio)

cpsievert:
Just FYI @hadley -- https://github.com/timelyportfolio/listviewer

hadley:
@cpsievert thanks - I had that bookmarked from twitter too.

PR #29, opened 2015-05-01
Package does not contain xml function

darrkj:
The README has code that does not work. Swap function xml out for read_xml.


hadley:
Thanks!

PR #37, opened 2015-05-29
converted configure script to autoconf

wrathematics:
I've converted the configure script over to use autoconf, to hopefully appease the issue with CRAN.  I'll be honest, I'm not sure what you were using that constituted a bash extension.  But the generated configure script looks agnostic enough, so hoping this helps.

Obviously make sure to test on Windows and Mac (I have access to neither at the moment); seems to work on the few differently configured Linux builds I have close by.  And I guess we'll hear from Travis in a moment.


wrathematics:
Doh, it's the $OSTYPE var.  I'll fix it.

PR #38, opened 2015-05-29
Removed bash extensions

wrathematics:
I converted the configure script over to autoconf and removed (I'm pretty sure) the bash extensions that got the package removed from CRAN.

Looking over it a bit more carefully, I think the $OSTYPE was the only bash extension.  So if you don't want to move to autoconf, you could pretty easily fix that in the original.  Won't hurt my feelings any.


jeroenooms:
Actually BDR has stated that `$(echo ...)` is a GNU extension.

wrathematics:
Oh, interesting.  Then you may need the conversion.

But $OSTYPE is definitely a BASH extension.  You were only using it for Macs where I guess you can reasonably assume it's there.  But when I switch to /bin/sh (dash on my system), it's not defined (not a mac though).

Anyway, the easiest way to check would be to just switch to a really basic shell and `/.configure`.

jeroenooms:
Are backticks really preferable over `$(...)`?

jeroenooms:
I really dislike shipping a 3000 line configure script which depends on the entire autoconf tool chain only to check if libxml2 headers are present. But if this will please CRAN I guess that is what we need to do.

Anyway I pushed some small fixes based on @wrathematics suggestions, and check for `pkg-config` as BDR mentioned in his email. Afaik this should run on any shell but I don't really have a way to test if it is truly bourne compatible. Lets try to submit and if it is still insufficient we will just switch to autoconf.

wrathematics:
Looks like all is well as the CRAN listing is back.  Closing.

Issue #53, opened 2015-10-04
Better missing handling for xml_find_one


hadley:
e.g. http://stackoverflow.com/questions/32885923

hadley:
See also https://github.com/hadley/rvest/issues/82

PR #57, opened 2015-11-06
Handle missing nodes

jimhester:
Return an empty list on missing and add default methods for all `xml_*`
methods.

I can add some tests once we have settled on the implementation, this is mainly a prototype for discussion.

Addresses https://github.com/hadley/xml2/issues/53, https://github.com/hadley/rvest/issues/82.


R/xml_find.R
```diff
@@ -93,6 +108,6 @@ xml_find_one.xml_nodeset <- function(x, xpath, ns = character()) {
   if (length(x) == 0)
     return(xml_nodeset())
 
-  nodes <- lapply(x, function(x) node_find_one(x$node, x$doc, xpath = xpath, nsMap = ns))
-  xml_nodeset(lapply(nodes, function(y) xml_node(y, x[[1]]$doc)))
+  xml_nodeset(lapply(x, function(x)
```
hadley: I think the checking for duplicate nodes needs to happen in `xml_nodeset()` because even for `xml_find_one` could return duplicate nodes (e.g. if your xpath was `.//x[1]`)

R/as_list.R
```diff
@@ -27,6 +27,11 @@ as_list <- function(x, ns = character(), ...) {
 }
 
 #' @export
+as_list.default <- function(x, ns = character(), ...) {
```
hadley: How much extra would it be to make this work with a specific class? I worry a little that a default method will make it easy for accidental mistakes to silently propagate

jimhester: Not much more work really, I just didn't know how we wanted to define that class, so using the default methods was a easy first draft.

hadley: Ok great :)

R/xml_children.R
```diff
@@ -81,7 +91,7 @@ xml_length.xml_node <- function(x, only_elements = TRUE) {
 #' @export
 xml_length.xml_nodeset <- function(x, only_elements = TRUE) {
   if (length(x) == 0)
-    return(integer)
+    return(integer(0))
```
jimhester: I figured you probably didn't want to return the `integer()` definition. :smile:

R/xml_find.R
jimhester: The duplication check is now moved to `xml_nodeset()` in https://github.com/jimhester/xml2/commit/ab8eaffda06a49a3dcc3bb6c93e2373af865d667

R/xml_write.R
```diff
@@ -17,6 +17,11 @@ write_xml <- function(x, file) {
 }
 
 #' @export
+write_xml.missing <- function(x, file) {
+  stop("Missing data cannot be written")
```
hadley: Please use `.call = FALSE`

jimhester:
Closing as I rolled this into #55

PR #59, opened 2015-11-13
Fix Coverage link, Add Build Status badge.

jimhester:
It was pointing at jimhester rather than hadley.

Also add a build status badge.


hadley:
Thanks!

PR #61, opened 2015-11-20
Changing configure script execution shell for portability.

heckendorfc:
Referencing /bin/bash is not portable and /bin/sh should be used instead. There are no bash-only features used in this script so there should be no problems with the change.


sanmai-NL:
Agree!

jimhester:
Thanks!

PR #69, opened 2015-12-17
* ENHANCE: Add test case causing a hard crash under 0.1.2.9000

sanmai-NL:
This test case crashes R with the following C++ exception: 
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```

### DIFF
--- a/tests/testthat/test-xml_find.R
+++ b/tests/testthat/test-xml_find.R
@@ -1,5 +1,12 @@
 context("xml_find")
 
+# Find all --------------------------------------------------------------------
+
+test_that("xml_find_all raises an error given an invalid XPath. ", {
+  x <- read_xml("<foo> <bar> text <baz/> </bar> </foo>")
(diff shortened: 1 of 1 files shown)


sanmai-NL:
@hadley: could you review this PR?

hadley:
That works just fine for me. If you're still having problems, I'd recommend that you file an issue.

sanmai-NL:
@hadley: Not sure what version you're testing with. The issue occurred back then with the version in the title, the test case is to prevent regressions. Is there a reason why you left the PR unmerged?

hadley:
It doesn't seem very useful to me.

sanmai-NL:
It could have prevented introduction of that bug earlier. This test is not all-important in isolation, but perhaps you should add more test cases like these for other functions. What kind of tests do you find important, actually? Would be good to take this into account and not waste time on writing tests you're not going to accept anyway.

PR #72, opened 2016-02-01
add only_leaves support to xml2::as_list

javierluraschi:
I spent some time finding ways to use `xml2` to import xml tabular data, not sure if I missed something obvious, but would like to share what I found...

In general, I wanted to avoid writing a big function to parse the XML tree to extract the data I was interested in. While writing this function is feasible, I was planning to use this code on the data import code generation module in RStudio; therefore, it would be really nice to generate something clean that is reusable.

I found a couple scenarios that work quite well.

For instance, for simple tabular XML:

```
<d>
    <r>
        <a>1</a>
        <b>1.1</b>
    </r>
    <r>
        <a>2</a>
        <b>1.2</b>
    </r>
</d>
```

This seems reasonable to me:

```
xml <- "<d><r><a>1</a><b>1.1</b></r><r><a>2</a><b>1.2</b></r></d>"
xml_list <- as_list(read_xml(xml))
View(matrix(unlist(xml_list), nrow = length(xml_list[[1]]), byrow = TRUE))
```

Also, for xml files with data nested in the hierarchy, this turned out to be easy to achieve with xpath:

```
xml <- "<q><c>caption</c><d><r><a>1</a><b>1.1</b></r><r><a>2</a><b>1.2</b></r></d></q>"
xml_list <- as_list(xml_find_one(read_xml(xml), "/q/d"))
View(matrix(unlist(xml_list), nrow = length(xml_list[[1]]), byrow = TRUE))
```

However, the first issue I hit was while loading formatted data:

```
xml <- "<d>\n\t<r><a>1</a><b>1.1</b></r>\n\t<r><a>2</a><b>1.2</b></r>\n</d>"
xml_list <- as_list(read_xml(xml))
View(matrix(unlist(xml_list), nrow = length(xml_list[[1]]), byrow = TRUE))
```

Hence, this PR to introduce an `only_leaves` parameter to filter out anything that is not a leaf element. However, I do wonder if I missed a simpler way to implement this functionality without changing as_list().

While this covers all nicely structured XML tables, there is another set of tables that this won't cover yet, for instance:
- Empty elements: `<d><r><a/><b>1.1</b></r><r><a>2</a><b>1.2</b></r></d>`
- Missing elements: `<d><r><b>1.1</b></r><r><a>2</a><b>1.2</b></r></d>`
- Unordered elements: `<d><r><a>1</a><b>1.1</b></r><r><b>1.2</b><a>2</a></r></d>`

Therefore, I'm not 100% sure it's worth taking a change to `as_list` since there might be other cases worth considering, it seems to me as if we wanted a whole new `as_data_table` function to handle all these cases. I'll be happy to make some extra time to help expand this PR over time, let me know what you think.

### DIFF
--- a/R/as_list.R
+++ b/R/as_list.R
@@ -15,6 +15,7 @@
 #' }
 #'
 #' @inheritParams xml_name
+#' @param only_leaves Should only content from leaf nodes be included?
 #' @param ... Needed for compatability with generic. Unused.
 #' @export
 #' @examples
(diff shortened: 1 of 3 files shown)


hadley:
My sense is that it's not worth spending much time on tabular xml files because hardly anyone uses them.

javierluraschi:
@hadley interesting... I'll follow up with you at some point to understand this better, closing this PR.

PR #75, opened 2016-02-17
Use a hash table rather than tree

jimhester:
Probably won't matter much either way in practice, but since we already
are linking against boost might as well use a hash table.


jimhester:
We need to delete the travis cache for this repository to fix the failures.

Issue #49, opened 2015-08-26
How to ignore "insignificant whitespace"?


jennybc:
I'm using the term "insignificant whitespace" as defined in [What You Need to Know About Whitespace in XML](http://www.oracle.com/technetwork/articles/wang-whitespace-092897.html):

> *Insignificant whitespace* is used when editing XML documents for readability. These whitespaces are typically not intended for inclusion in the delivery of the document.

Consider XML that has been formatted for human eyeballs. `xml2` can read it w/o error and well-formed XPATH expressions do what one expects. But once you use `as_list()` or try to write it back out with `write_xml()`, you learn there are problems with whitespace.

``` r
cd <- read_xml("http://www.xmlfiles.com/examples/cd_catalog.xml")
```

Targetted queries work fine:

``` r
xml_find_all(cd, ".//TITLE")
#> {xml_nodeset (26)}
#>  [1] <TITLE>Empire Burlesque</TITLE>
#>  [2] <TITLE>Hide your heart</TITLE>
#>  [3] <TITLE>Greatest Hits</TITLE>
#>  [4] <TITLE>Still got the blues</TITLE>
#>  [5] <TITLE>Eros</TITLE>
#>  [6] <TITLE>One night only</TITLE>
#>  [7] <TITLE>Sylvias Mother</TITLE>
#>  [8] <TITLE>Maggie May</TITLE>
#>  [9] <TITLE>Romanza</TITLE>
#> [10] <TITLE>When a man loves a woman</TITLE>
#> [11] <TITLE>Black angel</TITLE>
#> [12] <TITLE>1999 Grammy Nominees</TITLE>
#> [13] <TITLE>For the good times</TITLE>
#> [14] <TITLE>Big Willie style</TITLE>
#> [15] <TITLE>Tupelo Honey</TITLE>
#> [16] <TITLE>Soulsville</TITLE>
#> [17] <TITLE>The very best of</TITLE>
#> [18] <TITLE>Stop</TITLE>
#> [19] <TITLE>Bridge of Spies</TITLE>
#> [20] <TITLE>Private Dancer</TITLE>
#> ...
```

But `as_list()` reveals some problems

``` r
str(as_list(cd), max.level = 1, list.len = 5)
#> List of 53
#>  $   : chr "\n  "
#>  $ CD:List of 13
#>   .. [list output truncated]
#>  $   : chr "\n  "
#>  $ CD:List of 13
#>   .. [list output truncated]
#>  $   : chr "\n  "
[...]

PR #76, opened 2016-02-17
IN_PROGRESS: Mutating/Creating nodes

jimhester:
Clearly this work isn't done, but this PR will be a good place for discussion.


jimhester:
@hadley I think this is mostly in a good state now and modifying existing nodes has most of the features I think we want (with possible exception of setting the namespace of a node). I wrote a simple vignette to explain the functionality but you can also look at the tests to get a feel for the API.

Questions
- What should the modification/replacement functions do with `xml_missing` objects, throw an error or return them unchanged? If the latter should we issue a warning?
- How do we want to support creation of new nodes? Currently you can use `read_xml()` with character vector input and use the modification functions to insert the node into a tree. But this is clearly less efficient than creating the nodes directly.
- `xml_text()<-` uses [xmlNodeSetContentLen](http://www.xmlsoft.org/html/libxml-tree.html#xmlNodeSetContentLen) internally, which means that if it is called on a parent with non-text children they will be destroyed. This is the expected behavior for xmlNodeSetContentLen, but it could cause confusion or errors for users if they have objects pointing to the child nodes.

R/modify.R
```diff
@@ -0,0 +1,90 @@
...
+# 1. Assignment methods for xml_missing objects? Error, warning or identity
+
+#' Replace a node with another node
+#'
+#' @param x a document, node or nodeset.
+#' @param copy whether to copy the \code{value} before replacing. If this is \code{FALSE}
+#'   then the node will be moved from it's current location.
+#' @param value node or nodeset to replace with.
+#' @export
+`xml_replace<-` <- function(x, copy = TRUE, value) UseMethod("xml_replace<-")
```
hadley: I think these look better with `{}`

R/modify.R
```diff
@@ -0,0 +1,90 @@
...
+
+  # Need to wrap this in a list if a bare xml_node so it is recycled properly
+  if (inherits(value, "xml_node")) {
+    value <- list(value)
+  }
+
+  Map(xml_add_sibling, rev(x), rev(value), where, copy)
+}
+
+#' Add a child node
```
hadley: I think all three replacement methods could be documented together

R/xml_attr.R
```diff
@@ -95,3 +97,73 @@ xml_attrs.xml_node <- function(x, ns = character()) {
...
+#' @export
+`xml_attrs<-` <- function(x, ns = character(), value) UseMethod("xml_attrs<-")
+
+#' @export
+`xml_attrs<-.xml_node` <- function(x, ns = character(), value) {
+  if (!is_named_or_null(value)) {
+    stop("`value` must be a named character vector or `NULL`", call. = FALSE)
+  }
+
+  attrs <- names(value)
```
hadley: Need to check this is not NULL, and elements not "" or NA? (You can just throw an error if not)

```R
has_names <- function(x) {
  nms <- names(x)
  if (is.null(nms)) {
    rep_along(x, FALSE)
  } else {
    !(is.na(nms) | nms == "")
  }
}
```

hadley:
For creation, I'm imagining something like:

```R
xml_new_node("nodename", child1, child2, attr1 = "foo", attr3 = "bar")
```

I'm not sure how namespaces would work here. One option would be to have special name and attribute constructors:

```R
xml_new_node(nsName("url", "nodename"), child1, child2, nsAttr("attr1", "url", "foo"))
```

And instead of url could you use a prefix, if you also supplied the ns map.

hadley:
Need to check that we can solve http://stackoverflow.com/questions/32939229/creating-xml-in-r-with-namespaces/32941524#32941524 reasonably elegantly.

jimhester:
After a81f82ce70 you can now do the following to get the desired XML from the SO post.

Maybe it would be nicer to have the second argument to `xml_new_namespace()` be a character vector of namespaces to add, with the names as the prefixes.
``` r
d <- xml_new_document(xml_new_node("sld", version = "1.1.0", xml_new_node("layer", xml_new_node("Name"))))
sld <- xml_find_one(d, "//sld")
name <- xml_find_one(d, "//Name")
xml_new_namespace(sld, "http://www.o.net/sld")
xml_new_namespace(sld, "http://www.o.net/ogc", "ogc")
xml_new_namespace(sld, "http://www.o.net/se", "se")
xml_set_namespace(name, "se")
cat(as.character(d))
#> <?xml version="1.0"?>
#> <sld xmlns="http://www.o.net/sld" xmlns:ogc="http://www.o.net/ogc" xmlns:se="http://www.o.net/se" version="1.1.0"><layer><se:Name/></layer></sld>
```

hadley:
Yeah, I like the idea to add multiple namespaces at once.

Could we make something like this work?

```R
xml_new_document() %>% 
  xml_add_child("sld", version = "1.1.0") %>% 
  xml_add_namespace(
    "http://www.o.net/sld", 
    ogc = "http://www.o.net/ogc",
    se = "http://www.o.net/se"
  ) %>% 
  xml_add_child("layer") %>% 
  xml_add_child("se:Name")
```

hadley:
Or maybe even:

```R
xml_new_document() %>% 
  xml_add_child("sld", 
    version = "1.1.0"
    ns = "http://www.o.net/sld", 
    "ns:ogc" = "http://www.o.net/ogc",
    "ns:se" = "http://www.o.net/se"
  ) %>% 
  xml_add_child("layer") %>% 
  xml_add_child("se:Name")
```

(That would require more work on the backend to parse the attribute names and create ns objects where necessary)

jimhester:
Yeah I can get that working. The main issue is the current `xml_add_*` functions all expect the `value` argument to be a `xml_node`, so I either need to define a `xml_add_new_*` functions to create a new node with the specified name or have different behavior if `value` is a `character(1)`. The current implementation was the simplest path forward with the current API, but I see the appeal of the proposed interface.

I think a format like the following will be possible (`ns` needs to be `.ns` to avoid classes with an attribute named `ns`).
```r
xml_new_document() %>% 
  xml_add_child("sld", 
    version = "1.1.0"
    .ns = c("http://www.o.net/sld", 
      ogc = "http://www.o.net/ogc",
      se = "http://www.o.net/se")
  ) %>% 
  xml_add_child("layer") %>% 
  xml_add_child("se:Name")
```

hadley:
Doing S3 dispatch on character vs. `xmlNode` seems reasonable to me. But I'm not sure what using `.ns` buys us over `ns`, `ns:ogc` etc.

jimhester:
How do you create an attribute named 'ns' otherwise? The argument would
have to be named 'xmlns' to avoid a conflict.

hadley:
Oh yeah, I meant `xmlns` instead of `ns` everywhere (I forgot what ns attributes look like)

jimhester:
as of c2fcf36a7 you can now do

``` r
d1 <- xml_new_document(
  xml_new_node("sld",
    version = "1.1.0",
    xmlns = "http://www.o.net/sld",
    "xmlns:ogc" = "http://www.o.net/ogc",
    "xmlns:se" = "http://www.o.net/se") %>%
  xml_add_child(value = "layer") %>%
  xml_add_child(value = "se:Name"))
as.character(d1)
#> [1] "<?xml version=\"1.0\"?>\n<sld xmlns=\"http://www.o.net/sld\" xmlns:ogc=\"http://www.o.net/ogc\" xmlns:se=\"http://www.o.net/se\" version=\"1.1.0\"><layer/><se:Name/></sld>\n"
```

Using the exact syntax for `xml_new_document()` is tricky as `xml_document` objects expect to have a root node defined. I am still working on a good way to implement it.

jimhester:
As of 2c3c002 you can use the following

``` r
d1 <- xml_new_document()
xml_add_child(d1, "sld",
    version = "1.1.0",
    xmlns = "http://www.o.net/sld",
    "xmlns:ogc" = "http://www.o.net/ogc",
    "xmlns:se" = "http://www.o.net/se") %>%
  xml_add_child("layer") %>%
  xml_add_child("se:Name")
#> {xml_node}
#> <Name>
as.character(d1)
#> [1] "<?xml version=\"1.0\"?>\n<sld xmlns=\"http://www.o.net/sld\" xmlns:ogc=\"http://www.o.net/ogc\" xmlns:se=\"http://www.o.net/se\" version=\"1.1.0\"><layer><se:Name/></layer></sld>\n"
```

The only thing preventing you doing a full chain like the original example is that `xml_add_child()` returns the child node rather than the parent so you have way to access the whole document otherwise.

hadley:
Perfect!!

@jennybc how does this look to you?

jennybc:
I will try this out in the next day or two.

jennybc:
Is there an equivalent of [`XML::xmlNode()`](http://www.inside-r.org/packages/cran/xml/docs/xmlNode), i.e. a function to create a single new node *de novo*? I'm trying to translate something currently implemented with `XML`. I create a gazillion nodes as components of a list with `purrr::pmap()` and `XML::xmlNode()`, based on a data frame where each row provides info needed to populate one node. Then I add them to a (new) xml document *en masse* via [`XML::addChildren(kids = LIST_OF_GAZILLION_NODES)`](http://www.inside-r.org/packages/cran/xml/docs/addChildren).

It feels like `xml2::xml_add_child()` can play the role `XML::addChildren()`, once I figure out how to create my gazillion nodes as a `nodeset` (?). But how am I supposed to I do that?

Am I overlooking some suitably wrapped version of the unexported function `create_node()`?

jimhester:
@jennybc we removed support for creating free nodes, as it makes it impossible to assign a namespace to the nodes on creation.

For your use case can't you just `pmap()` directly with `xml2::xml_add_child()`, to create and add the nodes in one operation?

If you can point to your XML example code I can also take a look at how it would be converted.

jimhester:
@jennybc Can you do something like the following?

``` r
``` r
mtcars$name <- gsub(" ", "-", row.names(mtcars))
d <- xml_add_child(xml_new_document(), "cars")
add_child <- function(...) {
  x <- purrr::map(list(...), as.character)

  do.call(xml_add_child,
    c(list(.x = d, .value = x$name), x[names(x) != 'name']))
}
purrr::pwalk(mtcars, add_child)
d
#> {xml_document}
#> <cars>
#>  [1] <Mazda-RX4 mpg="21" cyl="6" disp="160" hp="110" drat="3.9" wt="2.62 ...
#>  [2] <Mazda-RX4-Wag mpg="21" cyl="6" disp="160" hp="110" drat="3.9" wt=" ...
#>  [3] <Datsun-710 mpg="22.8" cyl="4" disp="108" hp="93" drat="3.85" wt="2 ...
#>  [4] <Hornet-4-Drive mpg="21.4" cyl="6" disp="258" hp="110" drat="3.08"  ...
#>  [5] <Hornet-Sportabout mpg="18.7" cyl="8" disp="360" hp="175" drat="3.1 ...
#>  [6] <Valiant mpg="18.1" cyl="6" disp="225" hp="105" drat="2.76" wt="3.4 ...
#>  [7] <Duster-360 mpg="14.3" cyl="8" disp="360" hp="245" drat="3.21" wt=" ...
#>  [8] <Merc-240D mpg="24.4" cyl="4" disp="146.7" hp="62" drat="3.69" wt=" ...
#>  [9] <Merc-230 mpg="22.8" cyl="4" disp="140.8" hp="95" drat="3.92" wt="3 ...
#> [10] <Merc-280 mpg="19.2" cyl="6" disp="167.6" hp="123" drat="3.92" wt=" ...
#> [11] <Merc-280C mpg="17.8" cyl="6" disp="167.6" hp="123" drat="3.92" wt= ...
#> [12] <Merc-450SE mpg="16.4" cyl="8" disp="275.8" hp="180" drat="3.07" wt ...
#> [13] <Merc-450SL mpg="17.3" cyl="8" disp="275.8" hp="180" drat="3.07" wt ...
#> [14] <Merc-450SLC mpg="15.2" cyl="8" disp="275.8" hp="180" drat="3.07" w ...
#> [15] <Cadillac-Fleetwood mpg="10.4" cyl="8" disp="472" hp="205" drat="2. ...
#> [16] <Lincoln-Continental mpg="10.4" cyl="8" disp="460" hp="215" drat="3 ...
#> [17] <Chrysler-Imperial mpg="14.7" cyl="8" disp="440" hp="230" drat="3.2 ...
#> [18] <Fiat-128 mpg="32.4" cyl="4" disp="78.7" hp="66" drat="4.08" wt="2. ...
#> [19] <Honda-Civic mpg="30.4" cyl="4" disp="75.7" hp="52" drat="4.93" wt= ...
#> [20] <Toyota-Corolla mpg="33.9" cyl="4" disp="71.1" hp="65" drat="4.22"  ...
#> ...
```

jennybc:
I am trying to whittle my problem down to something I can either (a) solve or (b) inflict on you @jimhester.

jennybc:
I choose (b) let @jimhester figure it out. Sorry this is not very minimal, but the examples above and the tests, etc., really don't get into the bits that feel hard about my actual problem. So I present in its full glory.

https://gist.github.com/jennybc/37398c989852aa0c7abd9c5cea75e4f6

There's an R script, corresponding md, input as csv, and XML output.

jennybc:
Some features of the XML creation style of `XML` that seem to be missing in what's proposed for `xml2`:

  * The ability to create free nodes, which is actually encouraged in `XML`. I liked that. Then again, I didn't like `XML`'s [memory leak problems](http://www.google.com/search?q=r+xml+package+memory+leak) and I suppose those are probably related 😬.
  * The ability to add lots of children at once and to define children recursively.

I think anyone translating from `XML` code will need to know how they're supposed to do these things with `xml2`. Or what to do instead.

jimhester:
@jennybc This is how I would reproduce the XML output with the current code (63e5c1c).

```r
update_fodder <- read_csv("27_update-fodder.csv")
africa_cellsfeed <- "SOME_URL"

library(xml2)
d <- xml_new_document() %>%
  xml_add_child("feed",
    xmlns = "http://www.w3.org/2005/Atom",
    "xmlns:batch" = "http://schemas.google.com/gdata/batch",
    "xmlns:gs" = "http://schemas.google.com/spreadsheets/2006")

d %>% xml_add_child("id", africa_cellsfeed)
f_XML <- function(cell, cell_id, edit_link, row, col, update_value) {
  d %>%
    xml_add_child("entry") %>%
    xml_add_child("batch:id", cell) %>%
    xml_add_sibling("batch:operation", type = "update") %>%
    xml_add_sibling("id", cell_id) %>%
    xml_add_sibling("link", rel = "edit", type = "application/atom+xml", href = edit_link) %>%
    xml_add_sibling("gs:cell", row = as.character(row), col = as.character(col), inputValue = update_value)
}

update_fodder %>% pwalk(f_XML)

write_xml(d, "27_feed-node-xml2.xml")
```
The only difference from the XML output is the use of a [XML declaration](https://xmlwriter.net/xml_guide/xml_declaration.shtml), which while optional is usually recommended, libXML includes it by default when you create a new document object.

```diff
diff -ub 27_feed-node-XML.xml 27_feed-node-xml2.xml
--- 27_feed-node-XML.xml        2016-05-13 09:17:26.000000000 -0500
+++ 27_feed-node-xml2.xml       2016-05-13 10:34:13.000000000 -0500
@@ -1,3 +1,4 @@
+<?xml version="1.0"?>
 <feed xmlns="http://www.w3.org/2005/Atom" xmlns:batch="http://schemas.google.co
m/gdata/batch" xmlns:gs="http://schemas.google.com/spreadsheets/2006">
  <id>SOME_URL</id>
  <entry>
```

jennybc:
Yes that all works me (with a caveat below about whitespace).

As for the XML declaration, I will try sending it along. I have just been following these instructions to the letter and it does not include such a declaration (select "protocol", if it's not already selected, to see the target XML):

<https://developers.google.com/google-apps/spreadsheets/#updating_multiple_cells_with_a_batch_request>

Now that `xml2` writes XML with line breaks, it needs to consume them as well. I updated [my gist](https://gist.github.com/jennybc/37398c989852aa0c7abd9c5cea75e4f6#file-27_test-drive-new-xml2-md) to show the problem and I opened an issue about this last year (https://github.com/hadley/xml2/issues/49).

jennybc:
I used @jimhester's basic approach to swap `xml2` for `XML` in all instances of XML creation in my actual package. Also switched to `xml2` for XML mutation where I was previously using regular expressions. Was able to remove `XML` entirely and all tests pass.

jenga

PR #77, opened 2016-02-19
Use actual maps for NsMap

jimhester:
This commit is included in #76, however I wanted to open a separate PR as I wanted to know your thoughts on a few things in particular.

This now uses bidirectional maps to do the lookups instead of the linear search. My original attempt at this was using boost::bimap, however boost::bimap::multimap_of doesn't seem to be part of BH for some reason, so I had to do the bimap by hand.

This probably won't effect performance one way or the other in practice much as the number of namespaces in a document will likely be small, so maybe it is not worth worrying about...

I know you mentioned we are not able to use C++11 due to backwards compatibility concerns, is that still the case? If so what is your thought on the `BOOST_AUTO` usage in this PR?

### DIFF
--- a/src/xml2_utils.h
+++ b/src/xml2_utils.h
@@ -5,7 +5,7 @@
 #include <libxml/tree.h>
 #include <boost/shared_ptr.hpp>
 #include <boost/typeof/typeof.hpp>
-#include <unordered_map>
+#include <boost/unordered_map.hpp>
 
 inline xmlChar* asXmlChar(std::string x) {
   return (xmlChar*) x.c_str();
@@ -60,7 +60,7 @@ public:
 class NsMap {
 
   // We only store the index to avoid duplicating the data
-  typedef std::unordered_map<std::string, std::string> url2prefix_t;
+  typedef boost::unordered_map<std::string, std::string> url2prefix_t;
 
   url2prefix_t url2prefix;
   std::vector<std::string> order;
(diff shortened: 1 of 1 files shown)


jimhester:
After thinking about this a little more I realized the pointers used in https://github.com/jimhester/xml2/commit/c6283ea229b23f60a7497485ae301b2176962606 woudl be invalidated when the map is re-allocated upon later insertion.

https://github.com/hadley/xml2/commit/e59eb79329609313dc54e62ccc0fe9ed1f887e7d simply uses maps of indexes into the two vectors, so it will continue to work regardless of subsequent insertions.

jimhester:
Because https://github.com/jimhester/xml2/commit/8578f383ec994a34818b4ce9e8446c48d2032578 deplicates the data twice an alternative implementation is https://github.com/jimhester/xml2/commit/b0f43e4ac6a93095003adf33b5994123989c611a, which only uses a map from the url to prefix. If we don't care about preserving the input order we wouldn't have to duplicate any data, but presumably we do?

Also I submitted a PR to bh adding bimap (https://github.com/eddelbuettel/bh/pull/24) which Dirk has kindly merged, so we could use that here after the next BH release.

hadley:
I'd prefer to not use `BOOST_AUTO`, and then do a solid switch to C++11 when R 3.4 comes out.  

For other comments, I've forgotten enough about the context that you'll need to remind me, so probably easiest to do on our call on Wed

jimhester:
I thought the new windows gcc toolchain was coming out with R 3.3. Isn't that the main thing preventing us from using C++11/14 now? Or is there something else we need to wait for R 3.4 for?

Talking about xml2 on Wed sounds good to me, we also need to hash out how we want to R API to work.

hadley:
Sorry that was a typo/thinko - I meant 3.3.

PR #85, opened 2016-05-12
Turn on XML_PARSE_HUGE

### LINKED ISSUE #70: Excessive depth in document error causing issues for dependent package rNOMADS
I maintain the rNOMADS package, which relies on xml2 to parse large lists of web-hosted files.  Unfortunately, attempts to retrieve some file list fails with

    Error: Excessive depth in document: 256 use XML_PARSE_HUGE option [1]

I have spent a great deal of time attempting to reset this depth in order to permit my users to download the data they need.  However, I believe this depth is hard coded and thus can't be changed.  This is a serious issue for the proper functioning of my package.  The only work around I can provide is to ask my users to download a forked version of xml2 that is not available on CRAN:

    library(drat)
    drat:::add("shabbychef")
    install.packages('xml2'

Would it be possible to implement shabbychef's fix in the official xml2 repository, so I can easily make it available to my users?
To reproduce error:

    library(xml2)
    library(lubridate)

    date.today <- format(now(tz = "UTC"), "%Y%m%d")
    url <- paste0("http://nomads.ncep.noaa.gov/cgi-bin/filter_rap.pl?dir=%2Frap.", date.today)
    xml2::read_html(url)

jimhester:
Fixes #70

### DIFF
--- a/src/xml2_doc.cpp
+++ b/src/xml2_doc.cpp
@@ -21,7 +21,7 @@ XPtrDoc doc_parse_file(std::string path,
     pDoc = xmlReadFile(
       path.c_str(),
       encoding == "" ? NULL : encoding.c_str(),
-      0
+      XML_PARSE_HUGE
     );
   }
 
@@ -42,7 +42,7 @@ XPtrDoc doc_parse_raw(RawVector x, std::string encoding,
       Rf_length(x),
       base_url == "" ? NULL : base_url.c_str(),
       encoding == "" ? NULL : encoding.c_str(),
-      HTML_PARSE_RECOVER | HTML_PARSE_NOERROR
+      HTML_PARSE_RECOVER | HTML_PARSE_NOERROR | XML_PARSE_HUGE
     );
   } else {
     pDoc = xmlReadMemory(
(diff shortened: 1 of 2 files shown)


jimhester:
One benefit of the current default is if the XML is it can catch malformed XML and erroring rather than allocating a bunch of memory and crashing. It might be worth keeping the current behavior the default and allowing users to turn on XML_PARSE_HUGE if needed.

hadley:
I'd have a slight preference towards making this an argument

(also needs a news bullet)

src/xml2_doc.cpp
```diff
@@ -42,15 +42,15 @@ XPtrDoc doc_parse_raw(RawVector x, std::string encoding,
       Rf_length(x),
       base_url == "" ? NULL : base_url.c_str(),
       encoding == "" ? NULL : encoding.c_str(),
-      HTML_PARSE_RECOVER | HTML_PARSE_NOERROR
+      HTML_PARSE_RECOVER | HTML_PARSE_NOERROR | XML_PARSE_HUGE
```
hadley: Are you sure `XML_PARSE_HUGE` is ok?

PR #87, opened 2016-05-13
Print attributes of the root node

### LINKED ISSUE #7: Print attributes of root node
And other xml metadata? (like pi's?)

jimhester:
This works for printing attributes now, but not for namespaces.

libxml2 doesn't seem to have an exported function to print just attributes, or to print a node _without_ recursively printing its children, which is why I had to do this in R code.

Printing namespaces is going to take a little more work as we don't currently have code to retrieve the namespace for a specific node, just the document as a whole.

Fixes #7


hadley:
The namespaces are no longer attributes once the xml2 has been parsed, right?

jimhester:
Right they are their own namespace objects.

PR #88, opened 2016-05-17
libxml2 parsing options

### LINKED ISSUE #62: xml2 with deep tree (too many nested elements).
I would like to open an issue to allow for a large number of elements. 
It has been raised before here but closed by the original author. However, the issue still remains. 

Here is where the problem has been raised here for processing weather files ... 
https://github.com/dannycbowman/cageo-rnomads/issues/1

Here is some more on the issue ... 
http://stackoverflow.com/questions/33819103/parsing-small-web-page-with-xml2-throws-xml-parse-huge-error

Here is where the original request was put in as an issue for xml2...
https://github.com/hadley/xml2/issues/44

jimhester:
Continuation of #85 

Fixes #62, #49


DESCRIPTION
```diff
@@ -1,5 +1,5 @@
 Package: xml2
-Version: 0.1.2.9000
+Version: 0.1.2.9200
```
hadley: BTW I don't normally bother bumping this unless it's needed for a revdep.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # xml2 0.1.2.9000
 
+* `xml_read()` now drops blank text nodes, preserves behavior of XML package
```
hadley: Maybe "mimicking" behaviour of XML package?

man/read_xml.Rd
```diff
@@ -40,6 +42,8 @@ encoding directive, this allows you to supply a default.}
 
 \item{as_html}{Optionally parse an xml file as if it's html.}
 
+\item{allow_huge}{Remove size and node limits when parsing.}
```
hadley: Maybe mention the downsides?

DESCRIPTION
jimhester: It was bumped by the original upstream commit, but good to know!

jimhester:
I ended up exposing all of the libxml2 parsing options to the R side in https://github.com/hadley/xml2/pull/88/commits/3f25fb685212b8929473cc4e3f65d5f57c58906e. They can be specified as a character vector of options to set. The option descriptions come from http://www.xmlsoft.org/html/libxml-parser.html#xmlParserOption

hadley:
This looks great

Issue #35, opened 2015-05-25
Namespace aliases


jeroenooms:
In this example borrowed form the rversions package `lp1:` and `lp3:` namespace are aliases of `D:` however `xml_ns` does not list them at all, making it difficult to query the document.

```r
library(curl)
library(xml2)
h <- handle_setheaders(new_handle(customrequest = "PROPFIND"), Depth="1")
req <- curl_fetch_memory("http://svn.r-project.org/R/tags/", handle = h)
doc <- read_xml(rawToChar(req$content))
prop <- xml_find_all(doc, ".//D:propstat/D:prop", xml_ns(doc))

# does not work
xml_find_one(prop, ".//lp1:getetag", xml_ns(doc))

# works
xml_find_one(prop, ".//D:getetag", xml_ns(doc))
```

hadley:
Why does that make it hard to navigate the document? The goal of `xml_ns` is to provide a unique abbreviation for each namespace.

jeroenooms:
I found it confusing that the node in the xml is named `lp1:getetag` but searching for this node gives:

```
> xml_find_one(prop, ".//lp1:getetag", xml_ns(doc))
xmlXPathEval: evaluation failed
Error: No matches
In addition: Warning message:
In node_find_one(x$node, x$doc, xpath = xpath, nsMap = ns) :
  Undefined namespace prefix [1219]
```

Only by manually inspecting the xml I found that `lp1` is an alias of `D` in this case.

hadley:
That's how xml namespaces work though - the prefix is completely arbitrary, it's only the url that identifies it.  

But maybe `xml_ns()` shouldn't just return unique prefixes?  Would need some testing - I'm not sure if the libxml's xpath code will allow the same url to be mapped to multiple prefixes.

jimhester:
Having aliases to the same URI seems to work OK in limited testing. I think we should _not_ return only unique prefixes.

``` r
library(curl)
library(xml2)
h <- handle_setheaders(new_handle(customrequest = "PROPFIND"), Depth="1")
req <- curl_fetch_memory("http://svn.r-project.org/R/tags/", handle = h)
doc <- read_xml(rawToChar(req$content))
ns <- xml_ns(doc)

# Add the additional alias
ns[["lp1"]] <- "DAV:"

prop <- xml_find_all(doc, ".//D:propstat/D:prop", ns)
xml_find_one(prop, ".//lp1:getetag", ns)
#> {xml_nodeset (141)}
[...]

PR #89, opened 2016-05-18
Attach the namespace to the xml_document when it is created

### LINKED ISSUE #28: Feature request: attach namespace to an xml document
This is something you mentioned yourself here:

<https://github.com/hadley/xml2/issues/24#issuecomment-93810992>

From where I sit, this would be a very handy thing.

Related question: could the result of `xml_ns()` be attached to a new xml document by default? Or just available by default? Or could that behaviour be requestable?

Consider this example from [here](http://www.edankert.com/defaultnamespaces.html):

```
x <- read_xml('
<catalog xmlns="http://www.edankert.com/examples/">
  <cd>
    <artist>Sufjan Stevens</artist>
    <title>Illinois</title>
    <src>http://www.sufjan.com/</src>
  </cd>
  <cd>
    <artist>Stoat</artist>
    <title>Future come and get me</title>
    <src>http://www.stoatmusic.com/</src>
  </cd>
  <cd>
    <artist>The White Stripes</artist>
    <title>Get behind me satan</title>
    <src>http://www.whitestripes.com/</src>
  </cd>
</catalog>')
```

Due to the use of a default namespace, this does not work:

```
> xml_find_all(x, "//cd")
{xml_nodeset (0)}
```

So I inspect the namespace.

```
> xml_ns(x)
d1 <-> http://www.edankert.com/examples/
```

OK now I have hope I can use the `d1` prefix to specify default namespace. I'm usually happy to accept a default prefix like this.

```
> xml_find_all(x, "//d1:cd")
xmlXPathEval: evaluation failed
{xml_nodeset (0)}
Warning message:
In node_find_all(x$node, x$doc, xpath = xpath, nsMap = ns) :
  Undefined namespace prefix [1219]
```

Sad. But this does work:

```
> xml_find_all(x, "//d1:cd", xml_ns(x))
{xml_nodeset (3)}
[1] <cd>\n              <artist>Sufjan Stevens</artist>\n              <title>Illi ...
[2] <cd>\n              <artist>Stoat</artist>\n              <title>Future come a ...
[3] <cd>\n              <artist>The White Stripes</artist>\n              <title>G …
```

Could any of that be made easier?

jimhester:
Fixes #28 

Needs some tests still.

Also I feel this might be confusing if a XPath query works on the document, but then doesn't work on the xml_nodeset returned.

We could pass the namespace object down to the nodeset objects as well, which might be the way to go...


hadley:
We could also change the default to `xml_ns(document_root(x))` - then you would be stuck using the default prefixes, but that might not be so bad when combined with #35.

@jennybc do you have any use cases where you don't want to use the default prefixes?

jimhester:
Yeah I like a default of `xml_ns(xml_root(x))`, will change.

jennybc:
No, I almost always use the default prefixes.

I see that I do make some use of `xml2::xml_ns_rename()` to give a new alias to the default namespace. I think that's due to something annoying about the Sheets API?

I don't understand why attaching the namespace to the xml document means we're stuck with the default prefixes? Can we not still rename them or add new prefixes?

I just know that this is what every call to `xml_find_XXX()` I have ever written looks like:

```
xml2::xml_find_XXX(x, xpath, ns = xml2::xml_ns(x))
```

and I wish it was implied that, until I say otherwise, I want to query `x` with its own namespace.

hadley:
@jennybc what I'm thinking is that `xml_find_xxx` would get a default value of `ns = xml_ns(x)`. If you're happy with the default prefixes, that'd mean you could just do `xml_find_XXX(x, xpath)` everywhere. If you didn't like the default prefixes, you'd still need to do `xml_find_XXX(x, xpath, ns = my_ns)`. (In other words, this will make your life much easier if you rely on the default prefixes, but it won't get better if you're customising them). 

Does that sound reasonable? Or do you want a system for modifying the prefixes for a document once and for all? (that's possible but would be a decent amount of work so I'd rather avoid if the proposed compromise works for you)

R/xml_find.R
```diff
@@ -62,23 +62,23 @@
 #' ')
 #' xml_find_all(x, ".//f:doc")
 #' xml_find_all(x, ".//f:doc", xml_ns(x))
-xml_find_all <- function(x, xpath, ns = character()) {
+xml_find_all <- function(x, xpath, ns = xml_ns(xml_root(x))) {
```
hadley: I think this would be better as `ns = xml_ns(x)`, i.e. the logic for finding the root should be in `xml_ns()` (so we should probably make it generic)

jennybc:
The proposed solution makes sense and would eliminate a great deal of fussiness. That alone would be awesome.

I'm pretty sure the stuff in [this gist](https://gist.github.com/jennybc/bbe4de369e8d3c9621c2b43949223b3b) is not on the table here, but I'll link just in case. It definitely would require something along the lines of 'modifying the prefixes for a document once and for all'.

hadley:
Hmmmmm, @jimhester how hard would it be to write a function that simply removes all the default namespace specifications from the tree? 

(@jennybc I think that gist is slightly too simple since you're talking about a "a" default namespace when in fact there might be multiple, but I suspect that's unlikely to arise very often in practice. Simply stripping all default namespaces seems like a nice hack that will help most people).

jennybc:
Oh, I have never had the pleasure of multiple default namespaces. Something to look forward to! Does that change anything re: the safety or feasibility of removing them? One risk I thought of: people can puzzle themselves if they write this de-namespaced XML back out, when it should have the same namespaces as the original. E.g. mutating XML coming from / going back to an API.

That said, if there was a way to make default namespaces disappear, I would gladly sign some sort of waiver.

hadley:
Namespaces were designed to let you combine xml documents where (say) in one document `<board>` refers to a plank of wood, and in another `<board>` refers to the people running the company.  If you have a document that where "board" means different things in different parts of the document, you must use namespaces as using the name of the node is not enough (that's why xpath works the way it does - it's being excessively cautious so that you never accidentally confuse same-named nodes).  However, I don't think this actually crops up very often in practice (and you'd normally be able to work around it by specifying a parent node), so I think for 95% of uses, wiping out the namespaces would make life much easier with little risk of problems.

jimhester:
Yes this would be possible. There are functions to remove the namespace declaration node and then reconcile the rest of the namespaces on the tree.

If you wanted to include the namespace at the end you would have to add it back manually.

I will work on it!

jennybc:
@jimhester Pardon me if I misunderstand. You will remove all namespaces? Or just the default one? My frustration is focused only on the default one(s). The kind that has no associated prefix (yet requires user to create a prefix in order to write XPath queries).

jimhester:
@jennybc Well xml2 isn't going to do anything by default. But we should support modifying namespace definitions including removing them, so you will be able to remove the default namespace from a node if you would prefer that behavior.

jimhester:
As of https://github.com/hadley/xml2/pull/89/commits/c2e65a4a399b87b9a50eb7cd1938bf99182a07c3 you can now add and remove namespaces.

```r
x <- read_xml('
<catalog xmlns="http://www.edankert.com/examples/">
  <cd>
    <artist>Sufjan Stevens</artist>
    <title>Illinois</title>
    <src>http://www.sufjan.com/</src>
  </cd>
  <cd>
    <artist>Stoat</artist>
    <title>Future come and get me</title>
    <src>http://www.stoatmusic.com/</src>
  </cd>
  <cd>
    <artist>The White Stripes</artist>
    <title>Get behind me satan</title>
    <src>http://www.whitestripes.com/</src>
  </cd>
</catalog>')

old <- as.character(x)

xml_attr(x, "xmlns") <- NULL

xml_find_all(x, "//cd")
#> {xml_nodeset (3)}
#> [1] <cd>\n  <artist>Sufjan Stevens</artist>\n  <title>Illinois</title>\n ...
#> [2] <cd>\n  <artist>Stoat</artist>\n  <title>Future come and get me</tit ...
#> [3] <cd>\n  <artist>The White Stripes</artist>\n  <title>Get behind me s ...

xml_attr(x, "xmlns") <- "http://www.edankert.com/examples/"

new <- as.character(x)

identical(old, new)
#> [1] TRUE
```

There is an error on travis which only seems to occur when running R CMD check on a similar machine. I can reproduce it using a vagrant VM but not when running the tests interactively or using

hadley:
Could we maybe have `xml_strip_ns()` too?

jimhester:
Does `xml_string_ns()` strip only default namespaces or all namespaces and does it work only on the root node or does it need to search all children for namespace declarations as well?

hadley:
It would recursively strip default namespaces - I think that's the most useful case.

jimhester:
As of https://github.com/hadley/xml2/pull/89/commits/2a2ced54eabdf541365b6ea674a43616301eeb17 you can now use `xml_strip_ns()` to strip all of the default namespaces from a document.
``` r
  x <- read_xml(system.file(package = "xml2", "tests/testthat/ns-multiple-default.xml"))
  x
#> {xml_document}
#> <root>
#> [1] <doc1 xmlns="http://foo.com">\n  <bar/>\n</doc1>
#> [2] <doc2 xmlns="http://bar.com">\n  <bar/>\n</doc2>
  ns <- unclass(xml_ns(x))
  expect_equal(ns, c(d1 = "http://foo.com", d2 = "http://bar.com"))
  expect_equal(length(xml_find_all(x, "//bar")), 0)

  xml_strip_ns(x)
  x
#> {xml_document}
#> <root>
#> [1] <doc1>\n  <bar/>\n</doc1>
#> [2] <doc2>\n  <bar/>\n</doc2>
  ns <- unclass(xml_ns(x))

  expect_equivalent(ns, character())
  expect_equal(length(xml_find_all(x, "//bar")), 2)
```

R/xml_modify.R
```diff
@@ -185,3 +185,12 @@ xml_new_document <- function(version = "1.0") {
   doc <- doc_new(version)
   structure(list(doc = doc), class = "xml_document")
 }
+
+#' Strip the default namespaces from a document
+#'
+#' @inheritParams xml_name
```
hadley: Missing `@export`

hadley:
* Need to cross-reference `xml_ns_remove()` from the xpath functions
* `xml_ns_remove()` needs a couple of examples
* Needs bullet point in NEWS.md

jennybc:
When I use `xml2` from this PR, `read_xml()` prints all my XML to screen.

jimhester:
@jennybc can you give an example of the output you are seeing and what you were expecting to see?

jennybc:
@jimhester See [this gist](https://gist.github.com/jennybc/d0f27bdce2a2e54a8a1f6d2e2dd07590).

jimhester:
@jennybc This behavior should be gone with https://github.com/hadley/xml2/pull/89/commits/44c89cf3defc8c94c299a2afb871036cd7ffc657. I had enabled a debug flag to include the debug file to dump output for `str.xml_node`, but doing so also turns on node dumping results when you read a file.

Interestingly I could only reproduce your results on Windows, perhaps because it uses a newer version of libxml2 by default than I have on my OS X machine.

Anyway I removed that code, so this should now be fixed.

jennybc:
Yes, it's fixed! I upgraded to Xcode 7.3.1 earlier this month, in case that sheds any light.

jennybc:
The default namespace now comes along when you use `xml_attrs()`, which breaks some things for me. Is that how this is supposed to work? Should I open a new issue about this?

jennybc:
Example

``` r
library(xml2)
x <- read_xml('
  <root xmlns="http://default.com" count="80" uniqueCount="33">
    <item>one</item>
    <item>two</item>
  </root>')
```

With hadley/xml2@248e9fb

```r
xml_attrs(x)
#> count uniqueCount 
#> "80"        "33" 
as.integer(xml_attrs(x))
#> [1] 80 33
```

With hadley/xml2@c93169a

```r
xml_attrs(x)
#>                count          uniqueCount                xmlns 
#>                 "80"                 "33" "http://default.com"
as.integer(xml_attrs(x))
#> Warning: NAs introduced by coercion
#> [1] 80 33 NA
```

jimhester:
Returning the namespace definitions as an attribute is how it is supposed to be working.

hadley:
In other words, it was a bug that they weren't returned previously.

PR #92, opened 2016-05-27
Rename xml_find_one() to xml_find_first() and deprecate xml_find_one()

jimhester:


R/xml_find.R
```diff
@@ -203,3 +207,13 @@ xml_find_lgl.xml_nodeset <- function(x, xpath, ns = xml_ns(x)) {
...
    logical(0)
 }
+
+# Deprecated functions ----------------------------------------------------
+
+#' @rdname xml_find_all
+#' @usage NULL
+#' @export
+xml_find_one <- function(x, xpath, ns = xml_ns(x)) {
+  .Deprecated("xml_find_first")
```
hadley: The only downside of formally deprecating (i.e. instead of throwing a message) is that this will possible create new warnings for revdeps, so we'll need to do a release announcement at least two weeks prior to release. We probably don't need to do that otherwise because we're only adding new features.

jimhester: I'll run a revdep_check() now and see how many packages generate warnings, if it is a lot we can just use a `message()` instead of `.Deprecated()`.

jimhester: Looks like there are only 6 packages that call `xml_find_one()` on CRAN now, counts are number of search results (files) which contain calls to `xml_find_one()`.

``` r
library(gh)
res <- gh("GET /search/code", q = "user:cran xml_find_one")

library(purrr)
names <- map_chr(res$items, function(x) x$repository$name)
table(names)
#> names
#>   docxtractr googlesheets          oai    rversions          rvg 
#>            3            1            1            1           16 
#>      svglite         xml2 
#>            4            4
```

jimhester: Also the `xml_attrs()` change to output namespaces as well will likely break some packages, so we may need to do the release announcement regardless...

hadley: OTOH a message won't force anyone to notice the change. I can probably find time to fix svglite.  Maybe we give people a month?  That would make a CRAN release around May 27, which would line up nicely with useR.

PR #93, opened 2016-05-27
Add response methods for read_xml and read_html

jimhester:
Also convert read_html to a generic.

Perhaps the `need_package("xml")` is unnecessary, as how do you get a response object without having httr installed, but I put it in there to be safe.


R/utils.R
```diff
@@ -20,3 +20,13 @@ quote_str <- function(x, quote = "\"") {
 
   paste0(quote, x, quote)
 }
+
+is_installed <- function(pkg) {
+  system.file(package = pkg) != ""
```
hadley: Can we use `requireNamespace(quietly = TRUE)` instead?

jimhester: Sure (this is originally from https://github.com/hadley/httr/blob/master/R/utils.r#L35)

hadley: You should never trust that guy's code!

hadley:
LGTM

PR #94, opened 2016-05-27
xml_child() function to make selection easier

jimhester:
With the current implementation you can do the following

``` r
x <- read_xml("<x><y/><z><y/></z></x>")

xml_child(x)
#> {xml_node}
#> <y>
xml_child(x, 2)
#> {xml_node}
#> <z>
#> [1] <y/>
xml_child(x, "y")
#> {xml_node}
#> <y>
xml_child(x, "z")
#> {xml_node}
#> <z>
#> [1] <y/>
```
Some questions before I add docs/tests
- Is this a good function signature or were you thinking of something different?
- Do we also want this same functionality for `xml_sibling()` and `xml_parent()`?
- Do we want these functions to be generic with methods for xml_node and xml_nodeset?
- Do we want to modify the existing `xml_parent()` to use this same signature or keep it as-is?


R/xml_children.R
```diff
@@ -34,6 +34,20 @@ xml_children <- function(x) {
...
+xml_child <- function(x, search = 1, ns = xml_ns(x)) {
+  if (length(search) != 1) {
+    stop("`search` must be of length 1", call. = FALSE)
+  }
+
+  if (is.numeric(search)) {
+    xml_children(x)[[search]]
+  } else if (is.character(search)) {
+    xml_find_first(x, xpath = paste0("./", search), ns = ns)
+  }
```
hadley: Needs `else` that throws error

hadley:
1. Yes, that looks right to me.
1. I don't think so.
1. I don't think it needs to be since `xml_children()` is already generic.
1. Leave as is.

R/xml_children.R
```diff
@@ -4,11 +4,14 @@
...
 #' returns all parents up to the root. \code{xml_siblings} returns all nodes
-#' at the same level.
+#' at the same level. \code{xml_child} makes it easy to specify a specific
+#' child to return.
 #'
 #' @inheritParams xml_name
 #' @param only_elements For \code{xml_length}, should it count all children,
 #'   or just children that are elements (the default)?
+#' @param search For \code{xml_child}, either the child number to return (by
+#'   position), or the name of the child node to return.
```
hadley: "If there are multiple child nodes with the same name, the first will be returned"

jimhester:
@hadley anything else you think needs to change with this or should I merge it?

hadley:
LGTM

PR #96, opened 2016-06-02
Add information on deleting nodes and freeing memory in the vignette

### LINKED ISSUE #64: Before CRAN submission ensure svglite failure fixed

jimhester:
Fixes #64 

@hadley wanted to send a PR with this just because it is different than we discussed at the work week. Trying to use the XmlNodePtr finalizer wasn't working because it would get called _after_ the finalizers for XmlDocPtr, so the memory would already be freed when the node finalizer would be called.

Having an explicit `free=` argument allows users to free the memory if needed, with the understanding this lets them shoot off their foot if they are not careful. The default does _not_ free the memory when the nodes are unlinked.


vignettes/modification.Rmd
```diff
@@ -119,6 +126,29 @@ xml_add_child(t1, read_xml("<test/>"))
 x
 ```
 
+## Removing nodes ##
+The `xml_remove_node()` can be used to remove a node (and it's children) from a
```
hadley: `xml_remove()` ?

vignettes/modification.Rmd
```diff
@@ -119,6 +126,29 @@ xml_add_child(t1, read_xml("<test/>"))
...
+tree. The default behavior is to unlink the node from the tree, but does not
+free the memory for the node, so R objects pointing to the node are still
+valid. This allows code like the following to work without crashing R
```
hadley: I think you need to be a bit clearer about leaking memory here: "This means the default behaviour of `xml_remove()` is to leak memory, but it allows code like the following...`

vignettes/modification.Rmd
jimhester: It is currently named `xml_remove_node()` but I am happy to make is simpler!

vignettes/modification.Rmd
```diff
@@ -119,6 +126,29 @@ xml_add_child(t1, read_xml("<test/>"))
...
+rm(x1)
+gc()
+
+x2
+```
+
+Setting the argument `free = TRUE` when calling `xml_remove_node()`
+also frees the memory when the node is unlinked. In this case it is the
+responsibility of the user to remove all objects pointing to the node or its
+children, as they no longer point to valid memory.
```
hadley: And I think you should include an example showing how you should use `free = TRUE`.  Maybe mention that `xml_remove()` should always be called with `free = TRUE`, but you need to be careful not to have multiple R objects that point to the same object?

This case is pretty pernicious:

```R
 <- read_xml("<a><b /><b><b /></b></a>")
xml_find_all(x, "//b")
```

vignettes/modification.Rmd
jimhester: Technically I guess it is only a leak if there are no longer pointers to the memory which is not the case here but I will try to clarify.

jimhester:
Ok I tweaked the text with your suggestions, thanks for the review and let me know if you think anything needs to be clarified.

vignettes/modification.Rmd
```diff
@@ -119,6 +126,42 @@ xml_add_child(t1, read_xml("<test/>"))
...
+rm(x1)
+gc()
+
+x2
+```
+If you are not planning on referencing these nodes again this memory is wasted.
+Calling `xml_remove(free = TRUE)` will remove the nodes _and_ free the memory
+used to store them.  **Note** In this case _any_ node which previously pointed
+to the node or it's children will instead be pointing to free memory and may
+cause R to crash. It is the responsibility of the user to remove any objects
```
hadley: I'd make this a little friendlier: "xml2 can't figure this out for you, so it's your responsibility to remove..."

hadley:
Feel free to merge after resolving my last comment

Issue #102, opened 2016-06-12
ASAN memory leaks


jeroenooms:
ASAN says we are leaking memory. One place I narrowed down is in `xml_attr`:

```r
doc <- xml2::read_html('<a href="http://www.google.com" />')
nodes <- xml2::xml_find_all(doc, "//a[@href]")
links <- xml2::xml_attr(nodes, "href")
```

Gives:

```
> doc <- xml2::read_html('<a href="http://www.google.com" />')
> nodes <- xml2::xml_find_all(doc, "//a[@href]")
> links <- xml2::xml_attr(nodes, "href")
>
> q('no')

=================================================================
==22520==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 22 byte(s) in 1 object(s) allocated from:
    #0 0x7f678ff0bf4a in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x93f4a)
    #1 0x7f678293c7f8 in xmlStrndup__internal_alias /root/libxml2-2.9.4/xmlstring.c:45

SUMMARY: AddressSanitizer: 22 byte(s) leaked in 1 allocation(s).
```

The docs for `xmlGetProp` and `xmlGetNoNsProp` state *It's up to the caller to free the memory with xmlFree()*. Are we doing that somewhere?
 
There are others too, here is what leaks during the unit tests:

```
DONE ===========================================================================
> q("no")
=================================================================
==22774==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 46 byte(s) in 10 object(s) allocated from:
    #0 0x7f783dd43f4a in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x93f4a)
    #1 0x7f782f7567f8 in xmlStrndup__internal_alias /root/libxml2-2.9.4/xmlstring.c:45

Direct leak of 17 byte(s) in 2 object(s) allocated from:
    #0 0x7f783dd43f4a in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x93f4a)
    #1 0x7f782f713014 in xmlURIUnescapeString__internal_alias /root/libxml2-2.9.4/uri.c:1622

Indirect leak of 480 byte(s) in 4 object(s) allocated from:
    #0 0x7f783dd43f4a in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x93f4a)
[...]

PR #103, opened 2016-06-13
Fixes for memory leaks found with the LeakSanitizer

jimhester:
The diff looks bigger than it actually is, the main change in node_attr is assigning to a `xmlChar*` rather than a `const xmlChar*`. The former is freed by `Xml2String()` while the latter is not. xmlNsDefinition should return a `const xmlChar*` as previously as no string is allocated by libxml in that case.

`xmlURIUnescapeString` allocates a string if the target parameter is NULL. This is stated expliictly in the docs but is in the code (See https://github.com/GNOME/libxml2/blob/3eaedba1b64180668fdab7ad2eba549586017bf3/uri.c#L1620-L1627)

Those were all the direct leaks detected, I am still working on the indirect leaks.

Addresses #102

### DIFF
--- a/src/xml2_node.cpp
+++ b/src/xml2_node.cpp
@@ -75,31 +75,31 @@ SEXP node_attr(XPtrNode node, std::string name, CharacterVector missing,
     Rcpp::stop("`missing` should be length 1");
   SEXP missingVal = missing[0];
 
-  const xmlChar* string;
   if (name == "xmlns") {
-    string = xmlNsDefinition(node, NULL);
-  } else if (hasPrefix("xmlns:", name)) {
+    return CharacterVector(Xml2String(xmlNsDefinition(node, NULL)).asRString(missingVal));
+  }
+  if (hasPrefix("xmlns:", name)) {
     std::string prefix = name.substr(6);
-    string = xmlNsDefinition(node, asXmlChar(prefix));
+    return CharacterVector(Xml2String(xmlNsDefinition(node, asXmlChar(prefix))).asRString(missingVal));
+  }
+  xmlChar* string;
+  if (nsMap.size() == 0) {
+    string = xmlGetProp(node.get(), asXmlChar(name));
   } else {
-    if (nsMap.size() == 0) {
-      string = xmlGetProp(node.get(), asXmlChar(name));
+    size_t colon = name.find(":");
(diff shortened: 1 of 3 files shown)


jeroenooms:
You get a cleaner diff when adding `w=1` to the url: https://github.com/hadley/xml2/pull/103/files?w=1

src/xml2_node.cpp
```diff
@@ -75,35 +75,35 @@ SEXP node_attr(XPtrNode node, std::string name, CharacterVector missing,
...
+      std::string
+        prefix = name.substr(0, colon),
+               attr = name.substr(colon + 1, name.size() - 1);
 
-        std::string url = NsMap(nsMap).findUrl(prefix);
+      std::string url = NsMap(nsMap).findUrl(prefix);
 
-        string = xmlGetNsProp(node.get(), asXmlChar(attr), asXmlChar(url));
-      }
+      string = xmlGetNsProp(node.get(), asXmlChar(attr), asXmlChar(url));
```
jeroenooms: I think you're still leaking here? You need to call `xmlFree(string)` somewhere before returning...

src/xml2_node.cpp
```diff
@@ -75,35 +75,35 @@ SEXP node_attr(XPtrNode node, std::string name, CharacterVector missing,
...
     }
   }
 
-  return CharacterVector(Xml2String(string).asRString(missingVal));
+  return CharacterVector(Xml2String(string).asRString(missingVal));;
```
jeroenooms: Double semicolon looks like a typo?

src/xml2_node.cpp
jimhester: If a `xmlChar*` is passed to `Xml2String` it is freed in the `Xml2String` destructor (https://github.com/hadley/xml2/blob/master/src/xml2_utils.h#L26-L39), so this is free'd at L106.

src/xml2_node.cpp
jimhester: When is a double semicolon not a typo :) (odd for loops notwithstanding).

jimhester:
Ok I think we have all the leaks found in the tests now.

jeroenooms:
Looks good to me, thanks =)

PR #104, opened 2016-06-14
Use checked_get() in place of get() where possible

### LINKED ISSUE #101: Don't crash with null pointers
i.e when you've accidentally serialised and deserialised

jimhester:
This ensures we will generate an error rather than crashing on NULL
XPtrs (as happens when round tripping `saveRDS()/readRDS()`.

Fixes #101

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 # xml2 1.0.0
 
+* All C++ functions now use `checked_get()` instead of `get()` where possible,
+  so NULL XPtrs properly throw an error rather than crashing. (@jimhester,
+  #101, #104).
+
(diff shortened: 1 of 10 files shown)


jimhester:
@hadley PTAL

hadley:
LGTM. But why does `get()` even exist if there's `checked_get()`??

jimhester:
If you want to check for NULL yourself and not error.

```c
if (node.get() == NULL) {
  return CharacterVector()
}
```
ect.

jimhester:
Plus I there is slight overhead in the check you can avoid using `get()`

hadley:
Still, I think it would be better to have `get()` and `unsafe_get()` or similar.

hadley:
Can you add a bullet to news and merge please?

Issue #115, opened 2016-07-27
as_list.xml_node drops element names if any xml attributes exist


peterfoley:
Reproducible example:
```{r}
txt <- "<?xml version=\"1.0\" encoding=\"UTF-8\"?>
<level_0 xmlns=\"http://whatever.com\">
  <level_1a> 
    <item>item1a</item>
  </level_1a>
  <level_1b>
    <item>item1b</item>
  </level_1b>
</level_0>
"

packageVersion("xml2")
library(xml2)
parsed <- as_list(read_xml(gsub("\\n\\s*","",txt)))
names(parsed)
```

The issue appears to be line 58 in:
`attributes(out) <- as.list(attr)`
which overwrites the names that were previously stored by line 52.

leeper:
I am encountering the same issue with code that worked prior to xml2 v1.0.0 [but now fails](https://github.com/cloudyr/aws.s3/issues/60). I know this particular line of code was not changed in in the v1.0.0 release, but it appears to be interacting with other changes (probably in `xml_attrs()` methods).

Two difference changes could resolve this (for my use case):

1. Removing lines: https://github.com/hadley/xml2/blob/master/R/as_list.R#L58-L59
2. Replacing those lines with:

```R
 if (length(attr) > 0)
    attributes(out) <- c(attributes(out), as.list(attr))
```

The latter is probably better. Should I submit a PR?

peterfoley:
There's [already a PR](https://github.com/hadley/xml2/pull/116) basically done for this issue, just needs to be merged. The original issue was motivated by xml2 v1.0.0 breaking `cloudyr/aws.ec2`, so hopefully it'll also fix what was broken for `cloudyr/aws.s3`.

leeper:
Yes, it should. Thanks for that!

jimhester:
Sorry for the delay in merging, the current master should now be fixed.

PR #116, opened 2016-07-28
keep element names when xml node has attributes

peterfoley:
Fix for issue #115.

With this code, xml `names` attribute will still override the element names if both are specified. That's purely in the interest of maximum consistency with existing behavior, but I think the opposite would be more intuitive (element names win).

### DIFF
--- a/R/as_list.R
+++ b/R/as_list.R
@@ -61,8 +61,7 @@ as_list.xml_node <- function(x, ns = character(), ...) {
   if (length(attr) > 0) {
     # escape special names
     special <- names(attr) %in% c("class", "comment", "dim", "dimnames", "names", "row.names", "tsp")
-    if(any(special))
-      names(attr)[special] <- paste0(".", names(attr)[special])
+    names(attr)[special] <- paste0(".", names(attr)[special])
     attributes(out) <- c(list(names = names(out)), as.list(attr))
   }
 
--- a/R/as_list.R
+++ b/R/as_list.R
@@ -56,7 +56,7 @@ as_list.xml_node <- function(x, ns = character(), ...) {
   # Add xml attributes as R attributes
   attr <- xml_attrs(x, ns = ns)
   if (length(attr) > 0)
-    attributes(out) <- as.list(attr)
+    attributes(out) <- c(list(names=names(out)), as.list(attr))
 
   out
 }
--- a/R/as_list.R
+++ b/R/as_list.R
@@ -56,7 +56,7 @@ as_list.xml_node <- function(x, ns = character(), ...) {
   # Add xml attributes as R attributes
   attr <- xml_attrs(x, ns = ns)
   if (length(attr) > 0)
-    attributes(out) <- c(list(names=names(out)), as.list(attr))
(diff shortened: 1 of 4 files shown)


R/as_list.R
```diff
@@ -56,7 +56,7 @@ as_list.xml_node <- function(x, ns = character(), ...) {
   # Add xml attributes as R attributes
   attr <- xml_attrs(x, ns = ns)
   if (length(attr) > 0)
-    attributes(out) <- as.list(attr)
+    attributes(out) <- c(list(names=names(out)), as.list(attr))
```
jimhester: Could you put a spaces around the `=` operator here.

tests/testthat/test-as.list.R
```diff
@@ -20,3 +20,8 @@ test_that("cdata nodes become character vectors", {
 test_that("xml attributes become R attibutes", {
   expect_equal(list_xml("<x a='1' b='2'></x>"), structure(list(), a = "1", b = "2"))
 })
+
+test_that("xml names are preserved when attributes exist", {
+  expect_equal(list_xml("<x a='1' b='2'><y>3</y><z>4</z></x>"),
+               structure(list(y=list('3'),z=list('4')), names = c("y","z"), a = "1", b = "2"))
```
jimhester: Spaces around `=` here as well.

jimhester:
Maybe the best option would be to escape special attribute names (‘class’, ‘comment’, ‘dim’,
     ‘dimnames’, ‘names’, ‘row.names’ and ‘tsp’) with a leading `.`. They all have special meanings and restricted values, so setting some of them based on the XML values are likely to break people's code.

peterfoley:
Changes have been made, with basic tests and documentation.

R/as_list.R
```diff
@@ -55,8 +58,13 @@ as_list.xml_node <- function(x, ns = character(), ...) {
 
   # Add xml attributes as R attributes
   attr <- xml_attrs(x, ns = ns)
-  if (length(attr) > 0)
-    attributes(out) <- as.list(attr)
+  if (length(attr) > 0) {
+    # escape special names
+    special <- names(attr) %in% c("class", "comment", "dim", "dimnames", "names", "row.names", "tsp")
+    if(any(special))
```
jimhester: This conditional is unnecessary, the assignment will work without error regardless of the value of special

``` r
letters[logical(0)] <- 1
letters[FALSE] <- 1
letters[NULL] <- 1
letters
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j" "k" "l" "m" "n" "o" "p" "q"
#> [18] "r" "s" "t" "u" "v" "w" "x" "y" "z"
```

jimhester:
Could you add a note to NEWS.md referencing the issue # and thanking yourself ('@peterfoley')

peterfoley:
Is there anything else that I need to do? I'm somewhat new to public PR processes and just want to make sure it's not waiting on me.

jimhester:
Sorry for the delay, I think it looks good, thanks again!

PR #125, opened 2016-09-14
Add xml_prepend_child and xml_append_child methods.

sjp:
While working with `xml2` I found that I was often attempting to prepend a node to the children of another node.

This is a straightforward task in `XML` as `XML` exposes an `at` argument which controls the index at which the node is inserted. Achieving similar behaviour, at least for the purpose of prepending) was rather tedious in `xml2`.

I've written some code which makes prepending a node simple, and added `xml_append_child()` as a symmetric counterpart (just aliases `xml_add_child()`).

I don't think adding children at a specific index that is not the head or tail of a nodelist is common, but if this is the case, perhaps that could be added later as an argument to `xml_add_child()`.


hadley:
How about we just have one function called `xml_insert_child()`?

R/xml_modify.R
```diff
@@ -139,6 +142,49 @@ xml_add_child.xml_nodeset <- function(.x, .value, ..., .copy = TRUE) {
 
 #' @rdname xml_replace
 #' @export
+xml_prepend_child <- function(.x, .value, ..., .copy = TRUE) {
+  UseMethod("xml_prepend_child")
+}
+
+#' @export
+xml_prepend_child.xml_node <- function(.x, .value, ..., .copy = inherits(.value, "xml_node")) {
+  children <- xml_children(.x)
```
hadley: Should this be children or contents? What if you want to insert after a text node?

sjp: Good catch. I've updated to use `xml_contents` instead.

sjp:
`xml_insert_child()` would work.

Were you thinking of having an argument controlling position? Something like `.where = c("first", "last")`?

hadley:
@sjp yeah, exactly.

hadley:
Although maybe in that case it should just be an argument to `xml_add_child()` and we don't need any new functions?

jennybc:
> Although maybe in that case it should just be an argument to xml_add_child() and we don't need any new functions?

Maybe it could be consistent with `tibble::add_row(.data, ..., .before = NULL, .after = NULL)` and `tibble::add_col()`? Or will node insertion always be limited to adding to head or tail?

jimhester:
I really dislike having both a `before` and `after` argument, complicates the API and implementation for no real gain that I can see. You can always specify `after = 0` to append before the first element :-/

@jennybc You can use `xml_add_sibling()` to insert anywhere.

jimhester:
xml_add_child now has a `.where` argument, so this PR is no longer needed (https://github.com/hadley/xml2/commit/1856377a5d04060a1825cc2770e3a2157695e40d). Thank you again for sending it however, it motivated the change!

Issue #146, opened 2016-11-14
xml_integer() does not return integer


ddiez:
For example:
```
x <- read_xml("<plot><point x='1' y='2' /><point x='2' y='1' /></plot>")
xml_integer(xml_find_all(x, "//@x")) %>% is
[1] "numeric"        "vector"         "atomic"         "vectorORfactor"
```
The culprit might be in xml_text.R:
```
#' @export
xml_integer.xml_nodeset <- function(x) {
  vapply(x, xml_integer, numeric(1))
}
```
Where changing `numeric(1)` for `integer(1)` fixes it. It seem the unit test for xml_integer is passing because it uses expect_equal instead of expect_identical (but I am just getting familiar with testthat so I may be completely wrong on this):
```
test_that("xml_integer() returns an integer vector", {
  x <- read_xml("<plot><point x='1' y='2' /><point x='2' y='1' /></plot>")

  expect_equal(xml_integer(xml_find_all(x, "//@x")),
    c(1L, 2L))
})
```

PR #147, opened 2016-11-15
Fixed return value in call to vapply in xml_integer.xml_nodeset.

ddiez:
This fixes [#146](https://github.com/hadley/xml2/issues/146). I haven't changed the unit test (mentioned in the issue) but maybe should?.

### DIFF
--- a/R/xml_text.R
+++ b/R/xml_text.R
@@ -1,7 +1,7 @@
 #' Extract or modify the text
 #'
 #' \code{xml_text} returns a character vector, \code{xml_double} returns a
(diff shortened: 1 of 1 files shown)


jimhester:
You are right about the unit test, it should be using `expect_identical` in this case. Could you make that change, add a note to NEWS.md about the fix and thank yourself, then I can merge this.

jimhester:
Actually I am going to make the changes myself, no need to do so.

ddiez:
Great! Thanks. Thinking about the unit test recently I don't know what would be best practice here. Use expect_equal + expect_is (for integer) or expect_identical?

PR #153, opened 2016-12-10
Add support for xsl transformations

jeroenooms:
This adds support for `xls` transformations. The only downside is that this changes the system dependency on linux.


hadley:
What's the advantage of adding it here instead of in its own package?

jeroenooms:
If you would do a separate package, you need to export the headers defining `XPtrDoc` and `XPtrNode`, and the other package has to do `LinkingTo: xml2` to be able to access the contents behind the xml2 externalptr objects. 

If you do a separate package, both packages would statically link their own version of libxml2 on Windows. You need to make sure that both packages link to the same or abi compatible versions of `libxml2`.

So it's just simpler and it might be useful make xslt operations directly accessible from xml2. OTOH maybe it's not critical enough to be part of the core `xml2` stuff.

jimhester:
@jeroenooms I just pushed a branch which provides the XPtr definitions (https://github.com/hadley/xml2/tree/exported_XPtr) if you want to try it out with a separate package.

We could probably mess with the library paths to link against the same libxml2 library for both packages, but CRAN policies would likely frown upon it.

jeroenooms:
@jimhester OK that seems to work. However I am a little unsure about the global `libxml2` init and cleanup methods. The `xslt` pkg would call ` xmlInitParser();` and `xmlSetStructuredErrorFunc()` even if `xml2` has already been loaded, for the case that the package is statically linked (and hence needs to initiate it's own copy of libxml2)

```c
void R_init_xslt(DllInfo *info) {
    // Check that header and libs are compatible
    LIBXML_TEST_VERSION

    xmlInitParser();
    xmlSetStructuredErrorFunc(NULL, handleError);


    xsltInit();


    xsltSetGenericErrorFunc(NULL, (xmlGenericErrorFunc) handleXsltError);

    // load EXSLT
    exsltRegisterAll();
}
```

However if the package is dynamically linking to `libxml2` it will already have been initiated. I am not sure of re-initiating is harmless. Moreover, in case of dynamic linking, it will override the global `ErrorFunc` that was set by the `xml2` package.

Even more problematic are global cleanup functions, because another package might be using the same dynamic copy of `libxml2`. 

We saw the same [problem](https://github.com/davidgohel/gdtools/pull/33) in svglite/gdtools where one package calling a the fontconfig cleanup would cause a segfault if fontconfig was dynamically linked and other packages were using it as well.

gaborcsardi:
Taking a step back, it would be nice to have a good generic, sustainable solution to this problem. @jeroenooms is probably right that right now the simplest and safest solution is to merge the code into `xml2`. But I don't think this solution is sustainable in general.

It is easy to come up with solutions that work either with libs shared across packages (i.e. Linux), or private libs (i.e. CRAN windows or macOS). The question is, is there a good solution that works both ways?

jeroenooms:
I tested on Windows and Linux and Mac, and the exported-xptr solution seems to work. 

I don't quite understand what happens if you detach+unload the `xslt` package after it has overwritten the libxml `ErrorFunc` and then raise an error from `xml2`, however I can't get it to crash so I guess that's good.

jeroenooms:
Closing this in favor of the other solution: export `xml_types.h` so that we can make a separate package to perform xslt operations on `xml2` objects.

PR #158, opened 2016-12-20
Connection support for writing XML and HTML

### LINKED ISSUE #157: add support for connections to `write_xml`
`write_xml` accepts a file path as its second parameter. It could also accept a `connection` or even a broader list of outputs similar to `cat()`

jimhester:
This provides full connection support for XML and HTML output using the
xmlSave API (http://xmlsoft.org/html/libxml-xmlsave.html). It also
refactors the file output functions to use the xmlSave API as well,
which provides more formatting and encoding options than the previous
APIs.

Closes #157

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -31,3 +31,10 @@ need_package <- function(pkg) {
 
   stop("Please install ", pkg, " package", call. = FALSE)
 }
+
+# Format the choices for display in Rd
+rd_definition <- function(terms, defs) {
+  paste0("\\describe{\n",
+    paste0("  \\item{", terms, "}{", defs, "}", collapse = "\n"),
+  "\n}")
+}
--- a/R/utils.R
+++ b/R/utils.R
@@ -32,9 +32,11 @@ need_package <- function(pkg) {
   stop("Please install ", pkg, " package", call. = FALSE)
 }
 
-# Format the choices for display in Rd
-rd_definition <- function(terms, defs) {
+# Format the C bitwise flags for display in Rd. The input object is a named
+# integer vector with a 'description' character vector attribute that
+# corresponds to each flag.
+describe_options <- function(x) {
   paste0("\\describe{\n",
-    paste0("  \\item{", terms, "}{", defs, "}", collapse = "\n"),
+    paste0("  \\item{", names(x), "}{", attr(x, "description"), "}", collapse = "\n"),
   "\n}")
 }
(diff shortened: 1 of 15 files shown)


jimhester:
We could also convert the `doc_format*` and `node_format*` functions to use the xmlSave API (`xmlSaveToBuffer`), which would allow them to have an identical API to the write methods.

I wanted to be sure we were happy with this new API first before making that change.

R/utils.R
```diff
@@ -31,3 +31,10 @@ need_package <- function(pkg) {
 
   stop("Please install ", pkg, " package", call. = FALSE)
 }
+
+# Format the choices for display in Rd
+rd_definition <- function(terms, defs) {
```
hadley: Maybe make this take whatever `xml_save_options()` returns?

R/xml_write.R
```diff
@@ -1,81 +1,97 @@
...
     stop("Can only save length 1 node sets", call. = FALSE)
   }
 
-  node_write_xml(x[[1]]$node, x$doc, file, ...)
+  opts <- xml_save_options()
+  options  <- match.arg(options, names(opts), several.ok = TRUE)
+
+  if (inherits(con, "connection")) {
+    node_write_connection(x[[1]]$node, con, options = sum(opts[options]), encoding = encoding)
+  } else {
```
hadley: Maybe test explicitly for a length 1 character vector here?

src/xml2_output.cpp
```diff
@@ -61,6 +32,121 @@ CharacterVector doc_format_html(XPtrDoc x, bool format = true) {
...
+  write_ctxt ctxt;
+  ctxt.should_close = false;
+
+  ctxt.con = R_GetConnection(connection);
+  if (!ctxt.con->isopen) {
+    ctxt.con->open(ctxt.con);
+    ctxt.should_close = true;
+  }
+
+  xmlSaveCtxtPtr savectx = xmlSaveToIO(
```
hadley: WS needs some tweaking here

src/xml2_output.cpp
```diff
@@ -61,6 +32,121 @@ CharacterVector doc_format_html(XPtrDoc x, bool format = true) {
...
+  xmlSaveDoc(savectx, x.checked_get());
+  if (xmlSaveClose(savectx) == -1) {
+    stop("Error closing connection");
+  }
+}
+
+// [[Rcpp::export]]
+void node_write(XPtrNode x, std::string path, std::string encoding = "UTF-8", int options = 1) {
+  xmlSaveCtxtPtr savectx = xmlSaveToFilename(
+      path.c_str(),
```
hadley: We should be calling the C equivalent of `normalizePath()` here

hadley:
LGTM

PR #159, opened 2016-12-21
Implement xml_serialize and xml_unserialize methods

jimhester:
This isn't completely done, it still needs tests, but wanted to send it and see if there were any API or implementation changes needed.

### DIFF
--- a/R/xml_modify.R
+++ b/R/xml_modify.R
@@ -145,7 +145,7 @@ xml_add_child.xml_node <- function(.x, .value, ..., .where = length(xml_children
   num_children <- length(xml_children(.x))
 
   if (.where >= num_children) {
-    node_append_child(.x$node, node$node, .copy)
+    node_append_child(.x$node, node$node)
   } else if (.where == 0L) {
     node_prepend_sibling(xml_child(.x, search = 1)$node, node$node)
   } else {


R/as_xml_document.R
```diff
@@ -63,3 +63,17 @@ as_xml.list <- function(x, ..., missing = "elem") {
...
+}
+
+#' @export
+as_xml_document.xml_nodeset <- function(x, root, ...) {
+  doc <- xml_new_root(.value = root, ...)
+  for (i in seq_along(x)) {
+    xml_add_child(doc, x[[i]])
+  }
+  doc
+}
```
hadley: Probably should add `as_xml_document.xml_document()` for completeness (should just return `x`)

R/xml_serialize.R
```diff
@@ -0,0 +1,49 @@
...
+#'
+#' b <- xml_find_all(x, "//b")
+#' out <- xml_serialize(b, NULL)
+#' xml_unserialize(out)
+#' @export
+xml_serialize <- function(object, connection, ...) UseMethod("xml_serialize")
+
+#' @export
+xml_serialize.xml_document <- function(object, connection, ...) {
+  serialize(structure(as.character(object, ...), class = "xml_document"), connection)
```
hadley: These should probably use unique class names like `xml_serialised_document` (just so if you load it another way you don't get confused)

R/xml_serialize.R
```diff
@@ -0,0 +1,49 @@
...
+}
+
+#' @export
+xml_serialize.xml_document <- function(object, connection, ...) {
```
hadley: Check function name

R/xml_serialize.R
```diff
@@ -0,0 +1,49 @@
...
+}
+
+#' @rdname xml_serialize
+#' @export
+xml_unserialize <- function(connection, ...) {
+  object <- unserialize(connection)
+  if (inherits(object, "xml_nodeset")) {
+    x <- read_xml(unclass(object), ...)
+    return(xml_find_all(x, "/*/node()"))
+  }
```
hadley: I'd say drop the `return()` and add `else`s

hadley: Shouldn't the xpath just be `/node()` ? (if the intent is to handle nodesets with non-elements)

jimhester: I just want to select all the direct children of the root, this was taken from http://stackoverflow.com/a/691787/2055486, possibly `/node()` would work...

R/xml_serialize.R
jimhester: Yeah that is a fair point, it is safer that way

R/xml_serialize.R
```diff
@@ -17,33 +17,33 @@ xml_serialize <- function(object, connection, ...) UseMethod("xml_serialize")
...
+
+    # Select only the direct children of the root
+    xml_find_all(x, "/*/node()")
+  } else if (inherits(object, "xml_serialized_document")) {
     read_xml(unclass(object), ...)
-  }
-  if (inherits(object, "xml_node")) {
+  } else if (inherits(object, "xml_serialized_node")) {
     read_xml(unclass(object), ...)
   }
```
hadley: Should also have an `else` here for safety

PR #175, opened 2017-04-10
Add child node performance

heckendorfc:
I generate XML documents which can easily have nodes with >10k children. With the current logic in this package, there is no way to add child nodes without iterating through or creating R objects containing the complete list of children.

I restructured xml_add_child so that using `.where=0L` (prepend mode) will bypass these very slow function calls.

With the existing code, using `.where=0L` will produce an error if there are no children, rather than just insert the node as the first and only child. I assume this is a bug so I also included a fix for this in my changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# xml2 1.1.2
+
+* `xml_add_child()` now requires less resources to insert a node when called
+  with `.where = 0L` (@heckendorfc, #175).
+
 # xml2 1.1.1
 
 * This is a small point release addressing installation issues found with older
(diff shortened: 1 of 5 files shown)


jimhester:
This looks good, thank you for the PR! Could you add a test for the `.where = 0` case with no children to [tests/testthat/test-modify-xml.R](https://github.com/hadley/xml2/blob/master/tests/testthat/test-modify-xml.R)? Also please add a note to NEWS.md referencing this PR # and your GitHub handle. Thanks!

heckendorfc:
I can't reproduce that error I was talking about, so I'm going to guess I'm either thinking about something in `XML` rather than this package, or something I encountered during development. It's been a few days since I worked on this code.

The .where = 0L case should already be handled by the first test in:
 `"xml_add_child can insert anywhere in the child list"`

Recent commit adds NOTES message and fixes the improper text node handling the tests picked up.

jimhester:
Thanks!

Issue #191, opened 2017-10-06
Installation fails on macOS, with Xcode 9 installed


pkrog:
Under macOS Sierra 10.12.6 (16G29), using Xcode Version 9.0 (9A235), the following error is met when installing xml2 package:
```
clang++ -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/usr/local/opt/gettext/lib -L/usr/local/opt/readline/lib -L/usr/local/lib -L/usr/local/Cellar/r/3.4.2/lib/R/lib -L/usr/local/opt/gettext/lib -L/usr/local/opt/readline/lib -L/usr/local/lib -o xml2.so RcppExports.o connection.o xml2_doc.o xml2_init.o xml2_namespace.o xml2_node.o xml2_output.o xml2_schema.o xml2_url.o xml2_xpath.o -L/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk/usr/lib -lxml2 -lz -lpthread -licucore -lm -L/usr/local/Cellar/r/3.4.2/lib/R/lib -lR -lintl -Wl,-framework -Wl,CoreFoundation
ld: file not found: /usr/lib/system/libsystem_darwin.dylib for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [xml2.so] Error 1
ERROR: compilation failed for package ‘xml2’
* removing ‘/Users/pierrick/Library/R/3.4/library/xml2’
```

pigboysid:
I followed the instructions on ticket 191, which linked me to the post at https://stat.ethz.ch/pipermail/r-sig-mac/2016-September/012046.html:
* I copied `/usr/bin/xml2-config` to somewhere earlier in my path (in my case, `/Users/jcherry/bin`).
* I edited line 3 of the file: instead of `prefix=$(xcrun -show-sdk-path)/usr`, I changed it to `prefix=/usr`.

Following those two steps allowed `install.packages('xml2')` to complete successfully.

paul-riley:
Thanks @pigboysid! I just stumbled into this error and it fixed the linking issue I ran into.

kobeontheway:
I have the same problem ,you can try to reset the command line:
sudo xcode-select -s /Library/Developer/CommandLineTools

nazamy:
Thanks a bunch man!

[...]

PR #192, opened 2017-10-06
Solves installation issue with macOS Xcode 9.

pkrog:
Solves #191 .
The problem is that `xml2-config` returns a wrong/broken path for the library (`-L /Applications/.../MacOSX10.13.sdk/usr/lib`). See  <https://stat.ethz.ch/pipermail/r-sig-mac/2016-September/012046.html>, which refers to the same issue but in Xcode 8.
Correct the -L option given by `xml2-config` using sed: replace -L by -isysroot and correct path.

### DIFF
--- a/configure
+++ b/configure
@@ -5,6 +5,16 @@
 # INCLUDE_DIR and LIB_DIR manually via e.g:
 # R CMD INSTALL --configure-vars='INCLUDE_DIR=/.../include LIB_DIR=/.../lib'
 
+# Under macos, prefer pkg-config (if installed) over xml2-config, since the latter leads to a bug if Xcode 9 is installed:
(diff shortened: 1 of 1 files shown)


jimhester:
Thanks!

This has actually been a bug since Xcode 8 at least see https://stat.ethz.ch/pipermail/r-sig-mac/2016-September/012046.html.

I have a modified xml2-config in my bin directory with the line changed as suggested above, which is why I wasn't running into this myself.

Can you fix the added comments (which are not correct with your recent updates) and link to the email thread.

Also I would really _not_ recommend using homebrew R. Using the CRAN supplied R framework and CRAN package binaries in almost all cases. I realize it is unfortunate that homebrew R does not work with the CRAN binaries and you have to use an alternative installation procedure than for other software packages, but you are really better off doing so.

pkrog:
Hi, I've updated my initial comment, and will consider using R framework.
Regards

jimhester:
It doesn't look like you pushed your comment changes to the hotfix/xcode-9.0-compiling-issue branch.

pkrog:
Sorry, I misunderstood. Now it should be ok.

muschellij2:
Is this planned to be merged in the upcoming releases or is https://stat.ethz.ch/pipermail/r-sig-mac/2016-September/012046.html the recommended workaround?

jimhester:
Closed by https://github.com/r-lib/xml2/commit/404d5a559bce267595f7faa20caf6f6b5fafcef6

PR #247, opened 2019-02-12
add ... to xml_structure and descendents, closes #244

### LINKED ISSUE #244: FR: file argument for xml_structure
`xml_structure` is great, but the `cat` output can get quite long...

As opposed to `write_xml`, `file` argument adds flexibility to shut this argument on/off without changing much else while using `xml_structure`.

Would be straightforward to add since this argument can just be passed to `cat` directly, IIUC.

Happy to try out a PR if it sounds like a good idea.

MichaelChirico:
#244 

This is the simplest implementation but IMO not ideal:

 - A bit awkward about `sep`. Maybe make `sep` an argument as well? Seems the simplest way to catch duplicated `sep` at the outset. Otherwise will have to look at `match.call`? But I wasn't sure whether to only put a test for `sep` at the top (in `xml_structure` and `html_structure`?l `@export`ed functions? Whenever `cat` shows up?)

 - Seems a bit odd that you have to remember to supply `append = TRUE`... seems reasonable to expect a call to `xml_structure` to make only one file. And the current API means if your desired behavior is to overwrite an existing file with the full output of `xml_structure` (not a particularly odd behavior IMO), you're out of luck (or at least need to do some more manual stuff; I don't think a lazyeval trick with `append = file.exists(file)` will help since I think it will only evaluate once, not at each call?)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 * Windows: upgrade to libxml2 2.9.8
 
+* `xml_structure`, `html_structure` gain `...` argument which is passed to `cat`; useful e.g. for outputting to a file (@MichaelChirico, [#244](https://github.com/r-lib/xml2/issues/244))
+
 ## Bugfixes
 
 * Generic xml2 error are now forwarded as R errors. Previously these errors
--- a/R/xml_structure.R
+++ b/R/xml_structure.R
@@ -7,6 +7,7 @@
 #'
 #' @param x HTML/XML document (or part there of)
 #' @param indent Number of spaces to ident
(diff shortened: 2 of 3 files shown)


jimhester:
Thank you for working on the pull request! I ended up implementing this in a different way, exposing the `file` parameter rather than passing the dots (https://github.com/r-lib/xml2/commit/672e3f63b0201d9400237e80f31538e8102d7ccc) as that avoids the issues you described dealing with `sep` and `append`.

I did use the test from this PR, so thank you for including that!

Issue #232, opened 2018-09-28
Installation failure on macOS 10.14


jonclayden:
I apologise if this is something stupid, but I really can't see what I'm doing wrong here. I have updated this machine to macOS Mojave, using Homebrew-installed R. I can't get `xml2` to install, either with the system `libxml2` or the one from Homebrew:

```
$ R CMD INSTALL xml2_1.2.0.tar.gz 
* installing to library ‘/usr/local/lib/R/3.5/site-library’
* installing *source* package ‘xml2’ ...
** package ‘xml2’ successfully unpacked and MD5 sums checked
Found pkg-config cflags and libs!
Using PKG_CFLAGS=-I/usr/include/libxml2
Using PKG_LIBS=-L/usr/lib -lxml2 -lz -lpthread -licucore -lm
------------------------- ANTICONF ERROR ---------------------------
Configuration failed because libxml-2.0 was not found. Try installing:
 * deb: libxml2-dev (Debian, Ubuntu, etc)
 * rpm: libxml2-devel (Fedora, CentOS, RHEL)
 * csw: libxml2_dev (Solaris)
If libxml-2.0 is already installed, check that 'pkg-config' is in your
PATH and PKG_CONFIG_PATH contains a libxml-2.0.pc file. If pkg-config
is unavailable you can set INCLUDE_DIR and LIB_DIR manually via:
R CMD INSTALL --configure-vars='INCLUDE_DIR=... LIB_DIR=...'
--------------------------------------------------------------------
ERROR: configuration failed for package ‘xml2’
* removing ‘/usr/local/lib/R/3.5/site-library/xml2’
```

```
$ R CMD INSTALL --configure-vars='INCLUDE_DIR=/usr/local/opt/libxml2/include LIB_DIR=/usr/local/opt/libxml2/lib' xml2_1.2.0.tar.gz 
* installing to library ‘/usr/local/lib/R/3.5/site-library’
* installing *source* package ‘xml2’ ...
** package ‘xml2’ successfully unpacked and MD5 sums checked
Found INCLUDE_DIR and/or LIB_DIR!
Using PKG_CFLAGS=-I/usr/local/opt/libxml2/include 
Using PKG_LIBS=-L/usr/local/opt/libxml2/lib -lxml2
------------------------- ANTICONF ERROR ---------------------------
Configuration failed because libxml-2.0 was not found. Try installing:
 * deb: libxml2-dev (Debian, Ubuntu, etc)
[...]

PR #252, opened 2019-05-03
Using system libxml2 on Mojave

mys721tx:
`configure` prevents xml2 to find the system libxml2 on macOS Mojave. This PR adds the version number for Mojave.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,9 @@
 * Generic xml2 error are now forwarded as R errors. Previously these errors
   were output to stderr, so could not be suppressed (#209).
 
+* Stop rewritting `xml-config` on macOS El Capitan and later to remove
+  dependency on macOS SDK header package.(@mys721tx, #252)
+
 # xml2 1.2.0
 
 ## Breaking changes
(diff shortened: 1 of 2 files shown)


mys721tx:
This should allow detection of `libxml2` without the header package mentioned in #232.

jimhester:
Can you confirm this is true, I use Mojave, but I have the header installed, so it is not easy for me to verify.

```r
devtools::install_github("r-lib/xml2#252")
```

Will install using this PR and should be a good test if this is working properly (assuming you have not installed that header).

jimhester:
Can you please also add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

mys721tx:
The patch is unable to resolve the problem. Through further investigation, the problem we are facing are two folds:

1. The rewritten `PKGCONFIG_CFLAGS` is only available (`/usr/include/libxml2`) after installing the header package.
2. `pkg-config` installed from Homebrew default to `/usr/include/libxml2`

If a macOS user want to compile `xml2` on their machine, they will have a macOS SDK installed anyway. We should use the `xml-config` result directly, instead of rewriting it.

mys721tx:
I was able to build 3f53333 on a Mojave without the header package.

jimhester:
We can't just delete that logic completely, it is necessary for earlier versions of macOS and for instance CRAN's macOS machines are still running El Capitan.

mys721tx:
I just realized that de9781d has fixed this problem. However, it has not been released yet. @jimhester  Could you release bug fix version instead?

Issue #248, opened 2019-03-07
Register classes with setOldClass


nuest:
### Issue Description and Expected Result

It is not possible to use the class `xml_document` or `xml_node` in formal definitions of S4 classes.
It would be helpful for packages using S4 if they could properly define slots as being of a type provided by this package.

This should be possible with [`setOldClass`](https://www.rdocumentation.org/packages/methods/versions/3.5.2/topics/setOldClass).

(Let me know if you'd welcome a PR on this.)

### Reproducible Example

```r
library(xml2)

setClass("Testclass",
         representation(xml = "xml_document"),
         prototype = list(xml = xml2::xml_missing()),
         validity = function(object) {
           #...
           return(TRUE)
         }
)
```

```
Error in makePrototypeFromClassDef(properties, ClassDef, immediate, where) : 
  in making the prototype for class “Testclass” elements of the prototype failed to match the corresponding slot class: xml (class "xml_document" )
In addition: Warning message:
undefined slot classes in definition of "Testclass": xml(class "xml_document") 
```

<details>
<summary>Session Info</summary>

```r
devtools::session_info()

─ Session info ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value                       
 version  R version 3.5.2 (2018-12-20)
 os       Ubuntu 18.04.2 LTS          
 system   x86_64, linux-gnu           
 ui       RStudio                     
 language en_GB:en                    
 collate  en_GB.UTF-8                 
 ctype    en_GB.UTF-8                 
 tz       Europe/Berlin               
 date     2019-03-07                  

─ Packages ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 package     * version    date       lib source        
 assertthat    0.2.0      2017-04-11 [1] CRAN (R 3.5.1)
 backports     1.1.3      2018-12-14 [1] CRAN (R 3.5.1)
 bitops      * 1.0-6      2013-08-17 [1] CRAN (R 3.5.1)
 callr         3.1.1      2018-12-21 [1] CRAN (R 3.5.1)
 cli           1.0.1      2018-09-25 [1] CRAN (R 3.5.1)
 crayon        1.3.4      2017-09-16 [1] CRAN (R 3.5.1)
 desc          1.2.0      2018-05-01 [1] CRAN (R 3.5.1)
 devtools      2.0.1      2018-10-26 [1] CRAN (R 3.5.1)
[...]

PR #267, opened 2019-07-29
export classes and add docs

nuest:
see #248

`setOldClass` seems to require some manual definitions in roxygen, I followed https://stackoverflow.com/questions/7758748/documenting-setas-and-setoldclass-with-roxygen#7766762


jimhester:
Thanks! I ended up doing this in a somewhat simpler fashion. I will release an updated version of xml2 next week with this fix included, and you can submit sos4R at that time.

jimhester:
Oh, now I see you get a note if the classes are undocumented, so I will actually just use this PR afterall :)

nuest:
:smiley: 

I did the same thing: "Oh, that was quick to fix" and then `R CMD check` comes along and gives you a slap on the wrist. :man_shrugging:

nuest:
@jimhester Do you plan a bugfix release for this? Otherwise I'll ask CRAN to admit sos4R despite the warning..

jimhester:
Yes, I will send a release early next week.

PR #279, opened 2019-11-15
Add GPL-3 license and tidy-github docs

batpigandme:
For JOSS paper. cc @hadley


jimhester:
Hmm, not sure if we can change this to GPL-3, `GPL-2 (>=2)` is kind of vague, but seems more GPL-2 then three to me, changing it means we probably have to get approval from all the contributors.

hadley:
Oh yeah, we shouldn't change; @batpigandme just do what we did to the other packages in this situation

batpigandme:
Sorry, dropped the ball on this! Changed it back to GPL 2, @hadley!

jimhester:
Thanks!

PR #312, opened 2020-07-14
adding flatten argument for xml_find_all

### LINKED ISSUE #311: returning a list of nodesets
Hello, I have a use case in which I want to run `xml_find_all` on a nodeset, but have the result be a list of nodesets (one element for each node) instead of one flattened nodeset. In this use case, it's important that I know which result node came from which original node. In my testing it's way faster to have this functionality within the xml2 package, since it can directly call C in the lapply, than to call `xml_find_all` in a for or foreach loop. I can imagine a couple ways to do this.

1. Extend the `xml_find_all.xml_nodeset` method by adding an `unlist` argument. This would break the convention that "`xml_find_all` always returns a nodeset", but seems cleaner to me. Something like below:

```r
xml_find_all.xml_nodeset <- function(x, xpath, ns = xml_ns(x), unlist = TRUE) {
  if (length(x) == 0)
    return(xml_nodeset())

  if (isTRUE(unlist)) {
    nodes <- unlist(recursive = FALSE,
                    lapply(x, function(x)
                      .Call(xpath_search, x$node, x$doc, xpath, ns, Inf)))

    nodes <- xml_nodeset(nodes)

  } else {
    nodes <- lapply(x, function(x)
      xml_nodeset(.Call(xpath_search, x$node, x$doc, xpath, ns, Inf)))
  }

  return(nodes)
}
```

2. Add a separate function called `xml_find_all_list`, something like below:

```r
xml_find_all_list <- function(x, xpath, ns = xml_ns(x)) {
  if (length(x) == 0)
    return(xml_nodeset())

  lapply(x, function(x)
    xml_nodeset(.Call(xpath_search, x$node, x$doc, xpath, ns, Inf)))
}
```

I'm happy to make a PR, just let me know how to proceed. Thanks.

jakejh:
Fixes #311. Feedback welcome. This is my first time writing methods for S3 classes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # xml2 (development version)
 
+* `xml_find_all.xml_nodeset()` gains a `flatten` argument to control whether to return a single nodeset or a list of nodesets (#311)
+
 * `write_xml()` and `write_html()` now return NULL invisibly, as they did prior to version 1.3.0 (#307)
 
 # xml2 1.3.2
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # xml2 (development version)
 
-* `xml_find_all.xml_nodeset()` gains a `flatten` argument to control whether to return a single nodeset or a list of nodesets (#311)
+* `xml_find_all.xml_nodeset()` gains a `flatten` argument to control whether to return a single nodeset or a list of nodesets (#311, @jakejh)
 
 * `write_xml()` and `write_html()` now return NULL invisibly, as they did prior to version 1.3.0 (#307)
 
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # xml2 (development version)
 
+* `xml_find_all.xml_nodeset()` gains a `flatten` argument to control whether to return a single nodeset or a list of nodesets (#311)
```
jimhester, line 3:
```suggestion
* `xml_find_all.xml_nodeset()` gains a `flatten` argument to control whether to return a single nodeset or a list of nodesets (#311, @jakejh)
```

jimhester:
Looks great overall, thanks! Only change that is needed is adding your GitHub username to the NEWS, otherwise looks great as is.

jimhester:
Thanks!

PR #324, opened 2020-11-25
add a type check to xpath_search

MichaelChirico:
Ran into this when making an interactive typo:

```
Error in xml_find_all.xml_node(., pg, "//div[contains(text(), 'Results')]") : 
  STRING_ELT() can only be applied to a 'character vector', not a 'list'
```

The `STRING_ELT` part is likely to be puzzling, so we catch this mistake one line earlier instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 * `write_xml()` and `write_html()` now return NULL invisibly, as they did prior to version 1.3.0 (#307)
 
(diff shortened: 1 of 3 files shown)


jimhester:
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

MichaelChirico:
Done

jimhester:
Thank you!

PR #330, opened 2021-01-09
Add copy constructor and assignment operator for XPtr

MichaelChirico:
We ran into a segfault, apparently due to double-deletion on destruction of an XPtr (unfortunately I couldn't trace down a good repro of this).

Defining the copy constructor and assignment operator explicitly as here (see also: https://stackoverflow.com/a/4172724/3576984) solved the issue.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 * `xml_find_all.xml_node()` fails more informatively the `xpath` parameter is the wrong type (@michaelchirico)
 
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks, can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

MichaelChirico:
@jimhester done now, thanks

jimhester:
Thanks a bunch!

Issue #376, opened 2022-10-29
Trying to read an invalid .xml file makes it so that the file cannot be deleted


billdenney:
When trying to read an invalid .xml file, I cannot delete the.  I think that this is because the connection remains open.  If important, this happens on Windows (I've not tested other operating systems).

``` r
filename <- tempfile(fileext = ".xml")
bad_xml_file <- writeLines("<a>", con = filename)
file.remove(filename)
#> [1] TRUE

filename <- tempfile(fileext = ".xml")
bad_xml_file <- writeLines("<a>", con = filename)
xml2::read_xml(filename)
#> Error in read_xml.character(filename): EndTag: '</' not found [74]
file.remove(filename)
#> Warning in file.remove(filename): cannot remove file 'C:
#> \Users\wdenn\AppData\Local\Temp\RtmpQR65s3\file5e74463b35d8.xml', reason
#> 'Permission denied'
#> [1] FALSE
```

<sup>Created on 2022-10-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

billdenney:
My guess is that after 

https://github.com/r-lib/xml2/blob/ab2078976c9ab57116c7c9175632e7d19d45273f/R/xml_parse.R#L109

something like 

```r
on.exit({close(con)})
```

is needed.  But, I'm not sure.

PR #379, opened 2022-12-15
Add test for #376

billdenney:
This gives the (failing) test for #376.  The error occurs near https://github.com/r-lib/xml2/blob/ab2078976c9ab57116c7c9175632e7d19d45273f/R/xml_parse.R#L109

But, the actual error is within the C code of the function `doc_parse_raw`, and I don't know how to fix that.


hadley:
Note that this is a windows specific issue.

PR #391, opened 2023-05-09
Include Rinternals after vector

MichaelChirico:
Helps ensure R macros are resolved in the case of conflict (e.g. `<vector>` may start defining a `length()` macro that conflicts with R's)


hadley:
Oops, just fixed this problem myself since CRAN is now reporting on it.

MichaelChirico:
glad to hear it's tripping on CRAN too!

PR #399, opened 2023-08-22
Upkeep 2023

mgirlich:
This fixes part of #398.


mgirlich:
@hadley The PR looks quite big but as this is only upkeep most of the commits are pretty simple. If you prefer I could also break it down into multiple PRs.

hadley:
Looks great, thanks so much for doing this!

PR #400, opened 2023-08-23
Replace S3 dispatch by C implementation

### LINKED ISSUE #386: Trim all xml_text results from a nodeset at the same time
`xml_text` can take a long time when used with a large nodeset when `trim` is `TRUE`, and most of that time is spent calling `sub` twice for each node.  One experience I had was getting the text for a nodeset with about 4.5 million nodes, where `Rprof` showed that calls to `sub` made up 72% of the nearly two minutes spent in `xml_text`.

I'll happily make a pull request if you'd like

Here's a brief demo showing time saved:

```r
library(xml2)
library(microbenchmark)

blob <- paste0(c("<x>", rep("<y> Hi </y>", 1000), "</x>"), collapse = "")
tree <- read_xml(blob)
y_tags <- xml_find_all(tree, "//y")

xml_text_sub_after <- function(x, trim = FALSE) {
  res <- vapply(x, xml_text, trim = FALSE, FUN.VALUE = character(1))
  if (isTRUE(trim)) {
      res <- sub("^[[:space:] ]+", "", res)
      res <- sub("[[:space:] ]+$", "", res)
  }
  res
}

microbenchmark(
  as_is = xml_text(y_tags, trim = TRUE),
  proposed = xml_text_sub_after(y_tags, trim = TRUE),
  check = "identical"
)
# Unit: milliseconds
#      expr     min      lq      mean   median       uq     max neval
#     as_is 35.2209 35.9433 37.787917 36.64865 38.15000 58.4016   100
#  proposed  6.7719  6.9522  7.625734  7.16715  7.40195 16.2116   100
```

mgirlich:
Closes #386.

``` r
library(xml2)

data <- read_xml("http://aiweb.cs.washington.edu/research/projects/xmltk/xmldata/data/courses/reed.xml")
many_elts <- data |> 
  xml_contents() |> 
  xml_contents() |>
  xml_contents()

path <- xml2_example("cd_catalog.xml")
xml <- read_xml(path)
cds <- xml |> xml_children()
cd_contents <- cds |> xml_contents()

bench::mark(
  cds = xml_name(cds),
  cd_contents = xml_name(cd_contents),
  many_elts = xml_name(many_elts),
  check = FALSE
)

# DEV
#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           78.6µs   82.5µs   11139.     6.95KB     23.7
#> 2 cd_contents  447.2µs  480.9µs    1940.     1.27KB     21.2
#> 3 many_elts     28.2ms   30.1ms      33.3   65.72KB     42.8

# This PR
#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           3.52µs   3.87µs   214805.    2.66KB    21.5 
#> 2 cd_contents  13.03µs   14.1µs    61534.    1.27KB    18.5 
#> 3 many_elts     1.23ms   1.71ms      572.   65.72KB     8.36
```

<sup>Created on 2023-08-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details>

<summary>More benchmarks</summary>

``` r
library(xml2)

data <- read_xml("http://aiweb.cs.washington.edu/research/projects/xmltk/xmldata/data/courses/reed.xml")
many_elts <- data |> 
  xml_contents() |> 
  xml_contents() |>
  xml_contents()

path <- xml2_example("cd_catalog.xml")
xml <- read_xml(path)
cds <- xml |> xml_children()
cd_contents <- cds |> xml_contents()

bench::mark(
  cds = xml_name(cds),
  cd_contents = xml_name(cd_contents),
  many_elts = xml_name(many_elts),
  check = FALSE
)
#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds             66µs   78.6µs   11863.     6.95KB     23.8
#> 2 cd_contents  378.9µs  449.6µs    2069.     1.27KB     23.5
#> 3 many_elts     24.3ms   27.9ms      36.3   65.72KB     40.8

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           2.85µs   3.07µs   277778.    2.66KB     0   
#> 2 cd_contents  13.17µs  14.23µs    62456.    1.27KB    18.7 
#> 3 many_elts      1.1ms   1.63ms      616.   65.72KB     8.29

bench::mark(
  cds = xml_attr(cds, "test"),
  cd_contents = xml_attr(cd_contents, "test"),
  many_elts = xml_attr(many_elts, "test"),
  check = FALSE
)

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           80.7µs   95.9µs    9750.     8.66KB     25.5
#> 2 cd_contents  464.3µs  538.8µs    1779.     1.27KB     23.6
#> 3 many_elts     29.2ms   30.4ms      32.9   65.72KB    132.

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           2.77µs   3.01µs   292658.    3.44KB     29.3
#> 2 cd_contents   8.88µs   9.66µs    87308.    1.27KB     26.2
#> 3 many_elts   814.71µs   1.02ms      889.   65.72KB     10.4

bench::mark(
  cds = xml_attrs(cds),
  cd_contents = xml_attrs(cd_contents),
  many_elts = xml_attrs(many_elts),
  check = FALSE
)

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           71.6µs   83.8µs   11029.     6.48KB     25.6
#> 2 cd_contents  401.2µs  475.8µs    1997.     1.27KB     26.8
#> 3 many_elts     24.5ms   25.9ms      39.1   65.72KB    102.

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           3.54µs   4.39µs   194171.    3.02KB     38.8
#> 2 cd_contents  15.28µs  18.69µs    44680.    1.27KB     44.7
#> 3 many_elts     1.07ms   1.42ms      671.   65.72KB     29.9

bench::mark(
  cds = xml_text(cds),
  cd_contents = xml_text(cd_contents),
  many_elts = xml_text(many_elts),
  check = FALSE
)

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           94.8µs  110.1µs    7962.    11.24KB     22.1
#> 2 cd_contents  526.7µs  653.4µs    1241.     1.27KB     19.2
#> 3 many_elts     43.7ms   48.6ms      21.3   65.72KB     56.7

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           11.8µs  12.22µs    74143.    5.14KB     7.42
#> 2 cd_contents  37.14µs  42.75µs    22727.    1.27KB     9.09
#> 3 many_elts     1.94ms   2.36ms      418.   65.72KB     6.30

bench::mark(
  cds = xml_length(cds),
  cd_contents = xml_length(cd_contents),
  many_elts = xml_length(many_elts),
  check = FALSE
)

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           75.1µs   85.1µs    8996.     8.69KB     18.9
#> 2 cd_contents  417.2µs  475.9µs    1739.       672B     21.2
#> 3 many_elts     23.6ms   27.2ms      35.7   32.88KB     45.9

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds            1.9µs   2.06µs   418749.    2.41KB      0  
#> 2 cd_contents   6.13µs   6.71µs   121050.      672B     36.3
#> 3 many_elts   712.37µs 852.89µs     1075.   32.88KB     14.7

bench::mark(
  cds = xml_path(cds),
  cd_contents = xml_path(cd_contents),
  many_elts = xml_path(many_elts),
  check = FALSE
)

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           80.9µs   96.3µs    8871.     6.12KB     18.9
#> 2 cd_contents  499.1µs  614.8µs    1412.     1.27KB     14.6
#> 3 many_elts     56.2ms     67ms      15.6   65.72KB     26.0

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           15.7µs   18.5µs   52116.     2.46KB     5.21
#> 2 cd_contents  117.7µs  134.9µs    7305.     1.27KB     2.01
#> 3 many_elts     25.6ms     29ms      33.5   65.72KB     0

bench::mark(
  cds = xml_type(cds),
  cd_contents = xml_type(cd_contents),
  many_elts = xml_type(many_elts),
  check = FALSE
)

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           45.1µs   53.1µs   15496.      8.1KB     21.1
#> 2 cd_contents  232.4µs  281.6µs    2984.     1.92KB     21.3
#> 3 many_elts     16.1ms   17.8ms      52.9    98.6KB     26.5

#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 cds           1.85µs   2.22µs   360872.    5.73KB     36.1
#> 2 cd_contents   6.04µs    8.2µs   104980.    1.92KB     31.5
#> 3 many_elts   702.33µs    870µs     1076.    98.6KB     15.0
```

<sup>Created on 2023-08-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

</details>


mgirlich:
@hadley It would be great to get your opinion of whether this approach is fine for you. It kind of works against the dispatch system. But I'm also not quite sure whether the dispatch is really needed as there are only three classes (<xml_missing>, <xml_node> and <xml_nodeset>) and I don't think the user really needs to extend them.

The performance gains are quite big (around a factor of 20), which would be nice for e.g. the {paws} package (a package to work with AWS).

Overall, the code is pretty redundant so we could also use a macro or some other CPP techniques but I'm not so familiar with C/CPP to know about the potential downsides.

R/xml_text.R
```diff
@@ -109,7 +118,7 @@ xml_double.xml_node <- function(x) {
 
 #' @export
 xml_double.xml_nodeset <- function(x) {
-  vapply(x, xml_double, numeric(1))
+  as.numeric(xml_text(x))
```
hadley, line 121: Maybe this could just become the implementation for `xml_double()`? (And similarly for `xml_integer()`)

tests/testthat/test-xml_name.R
```diff
@@ -1,3 +1,13 @@
+test_that("xml_name() returns the name", {
+  x <- sample_nodeset()
```
hadley, line 2: IMO it would be better to copy and paste the contents of `sample_nodeset()` here; otherwise it's hard to verify that the test is correct.

src/xml2_node.cpp
```diff
@@ -37,6 +37,36 @@ extern "C" SEXP node_name(SEXP node_sxp, SEXP nsMap) {
...
+  BEGIN_CPP
+
+  int n = Rf_xlength(node_sxp);
+
+  SEXP out = PROTECT(Rf_allocVector(STRSXP, n));
+
+  for (int i = 0; i < n; ++i) {
+    SEXP node_sxp_i = VECTOR_ELT(node_sxp, i);
+
+    if (Rf_inherits(node_sxp_i, "xml_node")) {
```
hadley, line 51: I think this would be a little clearer/safer if you made a function that returns something enum-ish that you could then `switch` on.

hadley (approved):
I think switching from S3 dispatch to something custom at the C-level is fine, although as I noted I think providing a helper that makes it possible to use `switch` will make the implementation cleaner (and a little easier to extend in the unlikely event that is necessary).

I'm not too worried about the duplication, although the idiosyncratic mix of C and C++ is going to make maintenance generally harder. Possibly might be better to use cpp11 and some of the vector helpers that it provides.

inst/include/xml2_types.h
```diff
@@ -5,6 +5,24 @@
...
+    return(NodeType::node);
+  } else if (Rf_inherits(x, "xml_nodeset")) {
+    return(NodeType::nodeset);
+  } else if (Rf_inherits(x, "xml_missing")) {
+    return(NodeType::missing);
+  } else {
+    Rf_error("Unexpected node type");
+  }
+}
+
```
DavisVaughan, line 25: Is there a reason you put this here instead of, say, `xml2_utils.h` in `src/`?

If you put it here, it means any package that links against xml2 would also see this helper, and I don't think we want to make it public.

(for example, xslt by Jeroen https://github.com/ropensci/xslt/blob/56ceeeaee604fa872bb3f161d52a6b4b6a82796d/src/xslt.cpp#L12)

src/xml2_node.cpp
```diff
@@ -27,13 +27,57 @@ std::string nodeName(T* node, SEXP nsMap) {
...
+  switch(type) {
+  case NodeType::missing:
+    out = NA_STRING;
+    break;
+  case NodeType::node: {
+    SEXP node_sxp = VECTOR_ELT(x, 0);
+    XPtrNode node(node_sxp);
+
+    std::string name = nodeName(node.checked_get(), nsMap);
+    out = Rf_mkCharLenCE(name.c_str(), name.size(), CE_UTF8);
```
DavisVaughan, line 44: I suppose this has always been the case, but if we have an R error anywhere in here, then it will skip over any C++ destructors (like for XPtrNode, if it has one, and `std::string`) on the way out

I don't think there is a reason to try and improve this in this PR, but I will keep an eye out for potentially problematic cases as I continue my review

src/xml2_node.cpp
```diff
@@ -27,13 +27,57 @@ std::string nodeName(T* node, SEXP nsMap) {
...
+    break;
+  }
+  default: Rf_error("Unexpected node type");
```
DavisVaughan, line 47: Maybe extract out a `stop_unexpected_node_type()` helper that you can call, in case we ever want to tweak the message?

src/xml2_node.cpp
```diff
@@ -27,13 +27,57 @@ std::string nodeName(T* node, SEXP nsMap) {
...
-extern "C" SEXP node_name(SEXP node_sxp, SEXP nsMap) {
+extern "C" SEXP node_name(SEXP x, SEXP nsMap) {
   BEGIN_CPP
-  XPtrNode node(node_sxp);
+  NodeType type = getNodeType(x);
+
+  switch(type)
+  {
+  case NodeType::missing:
+  case NodeType::node   :
```
DavisVaughan, line 61:
```suggestion
  case NodeType::node:
```

We don't typically try and align based on the `:` in switches

src/xml2_node.cpp
```diff
@@ -82,15 +170,66 @@ const xmlChar* xmlNsDefinition(xmlNodePtr node, const xmlChar* lookup) {
   return NULL;
 }
 
+
+SEXP node_attr_impl(SEXP x,
+                    std::string name,
```
DavisVaughan, line 175:
```suggestion
                    const std::string& name,
```

Just to avoid copying in the loop

src/xml2_node.cpp
```diff
@@ -443,22 +613,70 @@ extern "C" SEXP node_children(SEXP node_sxp, SEXP only_node_sxp) {
...
+    }
+
+    SEXP out = PROTECT(Rf_allocVector(INTSXP, n));
+    int* p_out = INTEGER(out);
+
+    for (int i = 0; i < n; ++i) {
+      SEXP x_i = VECTOR_ELT(x, i);
+      int length_i = node_length_impl(x_i, only_node);
+      *p_out = length_i;
+      p_out++;
```
DavisVaughan, line 671:
```suggestion
      p_out[i] = length_i;
```

Seems a little clearer to me?

src/xml2_node.cpp
```diff
@@ -443,22 +613,70 @@ extern "C" SEXP node_children(SEXP node_sxp, SEXP only_node_sxp) {
...
+  case NodeType::missing:
+  case NodeType::node   :
+    return Rf_ScalarInteger(node_length_impl(x, only_node));
+    break;
+  case NodeType::nodeset: {
+    int n = Rf_xlength(x);
+
+    if (n == 0) {
+      return Rf_ScalarInteger(0);
+    }
```
DavisVaughan, line 662: I think everything is fine if you remove this. Generally i try and reserve early exits for cases where they are absolutely necessary (like if your algorithm requires >=2 elements or something, like most sorting algos)

DavisVaughan: Oh wait, I see what you are doing here. This seems odd. If the nodeset is length 0, shouldn't the output be length 0? i.e. `integer()`? In which case you could definitely just remove this then it would do the right thing.

Maybe this is some old behavior I don't know about

tests/testthat/test-xml_type.R
```diff
@@ -0,0 +1,18 @@
+test_that("multiplication works", {
```
DavisVaughan, line 1: MULTIPLICATION WORKS!

DavisVaughan (approved):

src/xml2_node.cpp
hadley: Yeah, we have an odd mix of C and C++ in this package which we should try and fix at some point.

inst/include/xml2_types.h
mgirlich: Ah, good point. I simply put it there because the file names seemed to suit well. But I didn't consider that this would be public then.

src/xml2_node.cpp
mgirlich: I also noticed this and found it quite weird which is why I opened an issue #404. I wanted to tackle that in a separate PR, forgot to mention this here.

tests/testthat/test-xml_type.R
mgirlich: 🎉  So happy to see THAT IT WORKS 🎉

hadley:
I'm guessing this is because while `inherits(x, "xml_node")` is still true, `Rf_inherits()` doesn't handle S4 classes.

PR #403, opened 2023-08-29
Remove unused dependencies

mgirlich:


hadley:
Probably added by `use_tidy_dependencies()`, but I agree that they're not important here.

PR #405, opened 2023-09-04
Use cpp11

mgirlich:
* how to deal with `memcpy()`?
* is the performance hit of the string handling relevant?
* `xml2_node.cpp` can be tackled after #400 is merged.


src/xml2_url.cpp
```diff
@@ -1,156 +1,133 @@
...
+  int n = x_sxp.size();
+  cpp11::writable::strings out(n);
+
+  const xmlChar* base_uri = to_xml_chr(base_sxp, "Base URL");
 
   for (int i = 0; i < n; ++i) {
-    const xmlChar* uri = (xmlChar*) Rf_translateCharUTF8(STRING_ELT(x_sxp, i));
-    SET_STRING_ELT(out, i, Xml2String(xmlBuildURI(uri, base_uri)).asRString());
+    const xmlChar* uri = (xmlChar*) Rf_translateCharUTF8(x_sxp[i]);
+    out[i] = Xml2String(xmlBuildURI(uri, base_uri)).asRString();
```
mgirlich, line 27: I guess the slow down due to cpp11 protection probably doesn't matter so much here

src/xml2_url.cpp
```diff
@@ -1,156 +1,133 @@
...
+cpp11::strings url_absolute_(cpp11::strings x_sxp, cpp11::strings base_sxp) {
+  int n = x_sxp.size();
+  cpp11::writable::strings out(n);
+
+  const xmlChar* base_uri = to_xml_chr(base_sxp, "Base URL");
 
   for (int i = 0; i < n; ++i) {
-    const xmlChar* uri = (xmlChar*) Rf_translateCharUTF8(STRING_ELT(x_sxp, i));
-    SET_STRING_ELT(out, i, Xml2String(xmlBuildURI(uri, base_uri)).asRString());
+    const xmlChar* uri = (xmlChar*) Rf_translateCharUTF8(x_sxp[i]);
```
mgirlich, line 26: Probably we can somehow leave away the `Rf_translateCharUTF8()` due to using cpp11.

src/xml2_url.cpp
```diff
@@ -1,156 +1,133 @@
...
 
-// [[export]]
-extern "C" SEXP url_absolute_(SEXP x_sxp, SEXP base_sxp) {
-  R_xlen_t n = Rf_xlength(x_sxp);
-  SEXP out = PROTECT(Rf_allocVector(STRSXP, n));
-
-  if (Rf_xlength(base_sxp) > 1) {
-    Rf_error("Base URL must be length 1");
+const xmlChar* to_xml_chr(cpp11::strings x, const char* arg) {
+  if (x.size() > 1) {
```
mgirlich, line 11: Might make more sense to check this in R.

src/connection.h
```diff
@@ -8,21 +10,7 @@
 #include <cstring>
 
 SEXP read_bin(SEXP con, size_t bytes = 64 * 1024);
-SEXP write_bin(SEXP data, SEXP con);
-
-inline SEXP R_GetConnection(SEXP con) { return con; }
```
mgirlich, line 13: I removed `R_GetConnection()` as it seemed unnecessary.

src/connection.h
```diff
@@ -8,21 +10,7 @@
...
-
-inline size_t R_ReadConnection(SEXP con, void* buf, size_t n) {
```
mgirlich, line 15: `R_ReadConnection()` wasn't used anywhere.

src/connection.h
```diff
@@ -8,21 +10,7 @@
...
-
-  memcpy(buf, RAW(res), size);
-
-  UNPROTECT(1);
-
-  return Rf_xlength(res);
-}
+cpp11::sexp write_bin(cpp11::sexp data, cpp11::sexp con);
 
 inline size_t R_WriteConnection(SEXP con, void* buf, size_t n) {
```
mgirlich, line 15: This uses `memcpy()` and I wasn't sure how to do this properly with cpp11.

DavisVaughan: I don't think there is an alternative to `memcpy()`, but you could allocate a cpp11 `raws` and then do `payload.sexp()` to get access to the SEXP you can call `RAW()` on. I don't think that is too awful, and at least lets cpp11 manage the memory

src/xml2_namespace.cpp
```diff
@@ -7,8 +9,8 @@
 #include "xml2_types.h"
 #include "xml2_utils.h"
 
-// [[export]]
-extern "C" SEXP unique_ns(SEXP ns) {
+[[cpp11::register]]
+cpp11::sexp unique_ns(SEXP ns) {
   BEGIN_CPP
```
DavisVaughan, line 14: Remove?

DavisVaughan: And in the other places below. Do we need this macro at all anymore? I assume if you go through cpp11 everywhere it will wrap it in try/catch for you

src/xml2_output.cpp
```diff
@@ -92,64 +91,56 @@ int xml_write_callback(SEXP con, const char * buffer, int len) {
...
-  BEGIN_CPP
+[[cpp11::register]]
+cpp11::sexp doc_write_connection(cpp11::sexp doc_sxp, cpp11::sexp connection, cpp11::strings encoding_sxp, cpp11::integers options_sxp) {
   XPtrDoc doc(doc_sxp);
-  const char* encoding = CHAR(STRING_ELT(encoding_sxp, 0));
-  int options = INTEGER(options_sxp)[0];
-
-  SEXP con = R_GetConnection(connection);
+  const char* encoding = cpp11::as_cpp<const char*>(encoding_sxp);
+  int options = options_sxp[0];
```
DavisVaughan, line 122: Or maybe `as_cpp<int>`?

src/xml2_url.cpp
DavisVaughan, line 11:
```suggestion
  if (x.size() != 1) {
```

Thinking about empty vectors too

src/xml2_url.cpp
DavisVaughan: Hmm, I don't _think_ cpp11 attempts to convert to UTF8 in this case
- When `x_sxp` is created as `cpp11::strings`, it just wraps the character vector SEXP with no translation on the elements
- When `x_sxp[i]` runs, it extracts the CHARSXP and turns it into a `r_string`, but that just calls this very simple constructor https://github.com/r-lib/cpp11/blob/500f642b4ea132ec8c168fc70a28e81e9510ece3/inst/include/cpp11/r_string.hpp#L17C10-L17C10
- On the way into `Rf_translateCharUTF8()` the `r_string` is automatically changed back to a SEXP by this operator, which also doesn't translate https://github.com/r-lib/cpp11/blob/500f642b4ea132ec8c168fc70a28e81e9510ece3/inst/include/cpp11/r_string.hpp#L22

It looks like if we went through an intermediate `std::string` it would translate, but what you currently have seems ok
https://github.com/r-lib/cpp11/blob/500f642b4ea132ec8c168fc70a28e81e9510ece3/inst/include/cpp11/r_string.hpp#L29

src/xml2_url.cpp
DavisVaughan: Yea if it isn't critical then I'd probably leave as is

src/xml2_url.cpp
```diff
@@ -1,156 +1,133 @@
...
-extern "C" SEXP url_escape_(SEXP x_sxp, SEXP reserved_sxp) {
-  R_xlen_t n = Rf_xlength(x_sxp);
-  SEXP out = PROTECT(Rf_allocVector(STRSXP, n));
-
-  if (Rf_xlength(reserved_sxp) != 1) {
-    Rf_error("`reserved` must be character vector of length 1");
-  }
+[[cpp11::register]]
+cpp11::strings url_escape_(cpp11::strings x_sxp, cpp11::strings reserved_sxp) {
+  int n = x_sxp.size();
```
DavisVaughan, line 106: The size method does actually still return a `R_xlen_t`, so it is probably best to keep that everywhere if it is easy
https://github.com/r-lib/cpp11/blob/500f642b4ea132ec8c168fc70a28e81e9510ece3/inst/include/cpp11/r_vector.hpp#L114

src/xml2_url.cpp
```diff
@@ -1,156 +1,133 @@
...
-    INTEGER(port)[i] = uri->port == 0 ? NA_INTEGER : uri->port;
-    SET_STRING_ELT(user, i, Rf_mkChar(uri->user == NULL ? "" : uri->user));
-    SET_STRING_ELT(path, i, Rf_mkChar(uri->path  == NULL ? "" : uri->path));
-    SET_STRING_ELT(fragment, i, Rf_mkChar(uri->fragment == NULL ? "" : uri->fragment));
+    scheme[i] = uri->scheme == NULL ? "" : uri->scheme;
+    server[i] = uri->server == NULL ? "" : uri->server;
+    port[i] = uri->port == 0 ? NA_INTEGER : uri->port;
+    user[i] = uri->user == NULL ? "" : uri->user;
+    path[i] = uri->path  == NULL ? "" : uri->path;
+    fragment[i] = uri->fragment == NULL ? "" : uri->fragment;
```
DavisVaughan, line 73: This one might get significantly slower, but again it doesn't _seem_ like it is used all that often? I couldn't really say

DavisVaughan (approved):

src/xml2_url.cpp
mgirlich: I also think it shouldn't be used that often.

src/xml2_namespace.cpp
mgirlich: @DavisVaughan If you could merge #400 then I could also finish xml2_node.cpp and I think then it should be safe to remove the macros.

mgirlich:
@DavisVaughan I now also migrated `xml2_node.cpp` to cpp11. But the performance hit from string handling really bites us now. It is nearly as slow as before replacing the S3 dispatch.

DavisVaughan:
Is it one or two functions in particular?

mgirlich:
I haven't looked into this in more detail but only checked `xml_name()` (see the benchmark in https://github.com/r-lib/xml2/pull/400). The easiest would probably be to check and carefully revert some of the changes and then wrap things with `cpp11::unwind_protect()`.

PR #406, opened 2023-09-26
Remove unused header

MichaelChirico:
Our tooling flagged this inclusion as unused, and looking around I think everything needed is pulled from <Rinternals.h> (`SEXP`) and <R_ext/Rdynload.h> (`R_CallMethodDef`, `DL_FUNC`, `R_registerRoutines`, `R_useDynamicSymbols`) so that seems correct.


hadley:
Thanks!

PR #412, opened 2023-10-31
New `xml_find_int()` to get integers from an XPath

### LINKED ISSUE #365: xml_find_int to return integer?
xml_find_num is a great tool for returning known-type output; any reason not to extend this to include a function returning known-integer output?

MichaelChirico:
Closes #365


tests/testthat/_snaps/xml_find.md
```diff
@@ -2,6 +2,10 @@
 
     Element at path `//z` must be a number, not a <xml_missing> object.
 
+# xml_find_int errors with non integer results
```
MichaelChirico, line 5: Looks like this only caught the first error. But this issue is also present in the other usages in this file -- I would relegate to a broader follow-up issue.

tests/testthat/test-xml_find.R
```diff
@@ -85,6 +85,26 @@ test_that("xml_find_num returns a numeric result", {
   expect_equal(xml_find_num(x, "1 div floor(-0)"), -Inf)
 })
 
+# Find int ---------------------------------------------------------------------
+test_that("xml_find_int errors with non integer results", {
+  x <- read_xml("<x><y/><y/></x>")
+  expect_snapshot_error({
```
hadley, line 91: We now prefer `expect_snapshot()` + `error = TRUE` so then you see the code and the error in the snapshot.

hadley (approved):

tests/testthat/test-xml_find.R
MichaelChirico: #414. Would you prefer fixing that first, or as a follow-up?

hadley: Fix it here, and then we'll update the other tests in a later PR.

PR #413, opened 2023-10-31
double-truncate to avoid inefficient encodeString() usage

### LINKED ISSUE #366: print.xml_nodeset very slow for huge document
Found in the XML representation of an edge case R file:

```r
library(xml2)
library(xmlparsedata)

p = parse("https://raw.githubusercontent.com/mwaldstein/edgarWebR/master/tests/cache/browse-edgar-11457c.R")
xml = read_xml(xml_parse_data(p))
```

Printing this is painfully slow:

```r
system.time(print(xml))
# {xml_document}
# <exprlist>
# [1] <expr line1="1" col1="1" line2="5944" col2="43" start="145" end="855979">\n  <expr line1="1" col1="1" line2="1" col2="9" start="145" end="153">\n    <SYMBOL_FUNCTION_CALL li ...
#    user  system elapsed 
#   2.906   0.048   2.958 
```

Took a brief look, it looks like `encodeString()` is the culprit:

```r
# ** debugging inside show_nodes() **
system.time(vapply(x, as.character, FUN.VALUE = character(1)))
#    user  system elapsed 
#   0.248   0.017   0.268 
system.time(encodeString(vapply(x, as.character, FUN.VALUE = character(1))))
#    user  system elapsed 
#   2.959   0.024   3.007
```

Is it possible to apply `substr()` twice -- once after `as.character()`, then again after `encodeString()`?

```
chr = vapply(x, as.character, FUN.VALUE = character(1))
nchar(chr)
# [1] 18965721
```

This is clearly already wayyy to wide (`width = 180` for me).

I _believe_ we can always just apply

```r
x %>%
  substring(1, n) %>%
  encodeString() %>%
  substring(1, n)
```

since the default behavior of `encodeString()` is to simply add `\` to non-printable characters, so it will just be a weakly wider version of the input.

Happy to file a PR if that sounds good.

MichaelChirico:
Closes #366.

There are some existing tests that were broken in as I was working on this, so I think those plus the new test are enough.

---

Re: the example raised in the issue, I see timing drop from ~1s to ~0.25s on my machine. Still noticeably slow, but a definite improvement. A quick profile shows we're going to need to drop to C++ to do any better.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # xml2 (development version)
 
 * Remove unused dependencies on glue, withr and lifecycle (@mgirlich).
+* `print()` is faster for very long `xml_nodeset` inputs (#366, @michaelchirico).
 
 # xml2 1.3.5
 
--- a/R/xml_nodeset.R
+++ b/R/xml_nodeset.R
@@ -38,6 +38,23 @@ as.character.xml_nodeset <- function(x, ...) {
   xml_nodeset(NextMethod())
 }
 
+#' Wrapper for encodeString() that takes width into consideration
+#'
+#' encodeString() is relatively expensive to run (see #366), so
+#'   avoid doing so to very wide inputs by first trimming inputs
+#'   to approximately the correct width, then encoding. A second
+#'   round of truncation occurs after encoding to account for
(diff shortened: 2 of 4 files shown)


hadley:
Thanks!

PR #417, opened 2023-11-20
Fix compilation with GCC 13

KNnut:
Tested with GCC 13.2.1 and libxml2 2.12.0.


hadley:
Thanks for figuring this out!!

PR #425, opened 2023-11-30
Handle inheritance for S4 objects

hadley:
@mgirlich could you please take a look too?

See https://github.com/r-lib/xml2/pull/400#issuecomment-1834643013 for details.


DavisVaughan (approved):
Yea I think this is the best way to do this without hurting performance in any way

mgirlich:
Also looks like a good solution to me.

PR #429, opened 2023-12-01
Use `expect_snapshot(error = TRUE)` over `expect_snapshot_error()`

### LINKED ISSUE #414: Use expect_snapshot()+error=TRUE over expect_snapshot_error() in tests
We now prefer `expect_snapshot()` + `error = TRUE` so then you see the code and the error in the snapshot.

_Originally posted by @hadley in https://github.com/r-lib/xml2/pull/412#discussion_r1380091891_

olivroy:
Fix #414


hadley:
Thank you!

PR #430, opened 2023-12-01
Lints to tests

olivroy:
Use `rlang::is_installed()`.

`expect_length()`, `expect_named()`, `expect_true()` in tests


hadley:
Thanks!

PR #431, opened 2023-12-01
Use cli messaging

### LINKED ISSUE #411: Use cli errors everywhere

olivroy:
Fix #411 

I also added full stops to error messages.

I enhanced the `parse_options()` message to use `{.val`, so it now uses `"` instead of `'`. I added a `error_call` argument. to get better error messages. (See snapshot.) + a link to [`read_html()` docs](https://xml2.r-lib.org/reference/read_xml.html#arguments) where the options are all documented.


hadley:
Thank you!

PR #445, opened 2024-08-25
GHA: test on macOS 13 + R-devel

gaborcsardi:
This is problematic currently.


hadley:
Is this still needed?

PR #462, opened 2025-07-23
Cite Alpine explicitly in configure log

MichaelChirico:
Useful to just read off in minimal environments like `docker.io/rhub/r-minimal:devel`; I trust `Alpine` is common enough to warrant surfacing here instead of just leaving in `etc`.


MichaelChirico:
count me in the minority then 😂

PR #479, opened 2026-06-20
perf: Make `xml_add_child()` append in constant time by default

krlmlr:
@jeroen: This is the cause for RSQLite with covr on GitHub Actions taking over 5 hours. Example run: https://github.com/r-dbi/RSQLite/actions/runs/27809553863

Cost driver: covr builds an XML tree by appending children. We can fix at the covr end, a fix here is likely to have a much more widespread impact. I can share a more detailed analysis and benchmarks if necessary.

Two independent Claude sessions came up with basically the same solution. The `.where = NULL` default is mine, Claude proposed `.where = Inf` . Reviewed manually. The `xml_length()` change only helps when a specific insertion point is passed. I'm not really sure about the logic of the "document" and "nodeset" methods, happy to add more tests.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # xml2 1.5.2
 
+* `xml_add_child()` without `.where` argument is now much faster. The default
+  is now `NULL` and still means "append after the last child".
+
 * `read_html()` now defaults to `encoding = "UTF-8"` to prevent double-encoding
   of UTF-8 content on Windows with codepage 65001 (#490).
 
--- a/R/xml_modify.R
+++ b/R/xml_modify.R
@@ -11,7 +11,10 @@
 #' @param .copy whether to copy the `.value` before replacing. If this is `FALSE`
 #'   then the node will be moved from it's current location.
 #' @param .where to add the new node, for `xml_add_child` the position
-#' after which to add, use `0` for the first child. For
+#' after which to add, use `0` for the first child. The default, `NULL`,
+#' appends after the last child without having to enumerate the existing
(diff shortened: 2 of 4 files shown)


krlmlr:
With the PR, the covr run shrinks to 6 minutes total, 2.5 minutes for covr: https://github.com/r-dbi/RSQLite/actions/runs/27885610548/job/82520936461?pr=479

jeroen:
@copilot does libxml2 have an efficient way to get the number of children from a node, rather than `length(xml_children(.x))`?

krlmlr:
This PR uses `xml_length()` . Claude claims there is no such way in libxml2, only head and tail can be accessed efficiently. Haven't verified that.

In a recent R:

``` r
FALSE || c(TRUE, FALSE, TRUE)
#> Error in `FALSE || c(TRUE, FALSE, TRUE)`:
#> ! 'length = 3' in coercion to 'logical(1)'
```

<sup>Created on 2026-06-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I suspect we'd need to establish argument checks at a large scale if we wanted to.

Should we run a round of revdepchecks just for this PR, or only before the release?

jeroen:
I guess internally libxml2 keeps the xml tree as a series of double-linked lists, so it has to iterate through the entire structure to find the length. I will merge it and trigger a revdep-check to see if there are any surprises.

jeroen:
I did a revdepcheck and there are no new issues with this change. 👍

jeroen:
This is now on CRAN on version 1.6.0.

krlmlr:
Thanks, great!

